PR #10, opened 2018-05-14
Features/combustion chamber stoich

fwitte:
I added the functionality for the stoichiometric combustion chamber. First cross check with basic combustion chamber showed good results. Perform tests in combination with other components, especially regarding convergence stability.


fwitte:
LUT-creation for custom fluids takes a long time, saving/reading to/from .csv should be added.

PR #12, opened 2018-06-20
Features/k a for design cases

pjhansen:
I am adding a ka feature for design case.


fwitte:
I will close this PR as the ability to specify kA-values for design-cases has been added in commit 79a1177.

PR #26, opened 2018-07-25
Features/speedy gonzales

fwitte:
I created a new feature branch for improvements on calculation speed. If you have any ideas or tips in order to speed up the calculation, you are welcome to post your ideas here! For now, I added an attribute to every component holding the incoming/outgoing connections, thus these do not have to be generated from the network object.


fwitte:
More improvements on calculation speed are always welcome!

PR #43, opened 2018-11-27
Features/negative mass flow

fwitte:
Allowing negative massflow for simple heat exchangers (pipes etc.).

### DIFF
--- a/tespy/components/characteristics.py
+++ b/tespy/components/characteristics.py
@@ -233,7 +233,8 @@ class characteristics:
             x['QLOSS'] = np.array([0.50, 0.7500, 0.90, 1.000])
             y['QLOSS'] = np.array([0.32, 0.3067, 0.30, 0.295])
 
-        elif self.comp == 'heat exchanger' or self.comp == 'desuperheater':
+        elif (self.comp == 'heat exchanger' or self.comp == 'desuperheater' or
+              self.comp == 'pipe' or self.comp == 'heat exchanger simple'):
 
             x['EVA_HOT'] = np.array(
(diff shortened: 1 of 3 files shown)


fwitte:
I added the necessary equations for the node. This component is able to switch back and forth between merge and splitter, depending on the mass flow directions. I will upload a running example of a very small pipe network with a loop, where the mass flow is distributed according to the pressure losses along two different paths on the [tespy_examples](https://github.com/fwitte/tespy_examples/tree/master/pipe_network) repository. Equations for simple heat exchangers/pipes have been adapted to usage with negative mass flow, already.

Beware of the following problems/difficulties:

- If you use a node and the flow direction changes within solution process, there must be a second node to cover up for the gain/loss of equations. Take a node with one incoming and two outgoing connections initially, for example. The enthalpy and fluid composition will be the same for the two outgoing mass flows as for the incoming one, thus using 2 * (1 (enthalpy) + n (number of fluids in network)) equations. If the flow direction changes, the fluid composition and enthalpy of the firstly outgoing and now incoming mass flow are not defined, thus missing one set of 1 + n equations. If you have two nodes connected to each other, the necessary flow direction inversion on the second node will cover up for the loss of equations.

- In process of flow direction change, if the flow direction into a sink or from a source is inverted, this will cause a similar problem. E. g. take a sink, where the outgoing fluid composition and enthalpy are to be determined by an adjacent node. If the mass flow direction changes, enthalpy and fluid composition are now unknown and cannot be determined by another node. This will lead to a linear dependency and render the problem unsolvable.

PR #57, opened 2019-02-05
solve the filename conflict in Windows by using tespy_fluid feature

ShuangChen88:
Dear Francesco,

Here i make a pull request for solving the filename conflict  by using tespy_fluid feature in Windows.

Based on my own understanding, i have changed the save and load path in TESPy for tespy_fluid from ‘TESPy::alias' to 'TESPy_alias'. I have tested it with the online documentation example and could achieve a successful save and load process. But i'm not sure if this change will make conflict in Linux system. So i would like to ask you if you could look into the code change and have a check for it. 

Best regards,
Shuang

### DIFF
--- a/tespy/tools/helpers.py
+++ b/tespy/tools/helpers.py
@@ -634,7 +634,8 @@ class tespy_fluid:
             Lookup value (enthalpy, entropy, density or viscosity)
         """
         df = pd.DataFrame(y, columns=x2, index=x1)
-        path = './LUT/' + self.alias + '/'
+        alias = self.alias.replace('::','_')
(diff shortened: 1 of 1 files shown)


fwitte:
Hi @ShuangChen88,
thank you for your effort. Since the 0.1.0 release, the code is be tested automatically. If you open a PR a background-check is performed by the travis-ci. You can click on the details to see the results of the test, if any errors occur.

resolve #57

ShuangChen88:
Sorry for closing this pull request because I think there is still problem with this change. I will check it again in my example.

PR #68, opened 2019-05-13
features/windows_mac_tests

fwitte:
Adding tests for windows and mac.


fwitte:
Removed the windows test for now, as it does not work at the moment. Will look into that after the release.

PR #79, opened 2019-07-16
Make combustion more flexible

### LINKED ISSUE #76: No/Wrong errormessage, if the network has no connections
If no connections have been added to the network, an AttributeError is thrown: `AttributeError: 'DataFrame' object has no attribute 'tolist'`. The message should be more clear, to indicate, that the user forgot to add connections to his network.

### LINKED ISSUE #77: Add flexibility to the combustion chamber and cogeneration unit
The fuel should be detected automatically from a given list of fuels. The should not need to attribute the fuel. This way, it is possible to use mixtures of different fuels on these components, e. g. for representing natural gas combustion.

fwitte:
The combustion chamber and cogeneration unit should work with a mix of different fuels and not only one specific fuel. The keyword argument fuel will be deprecated as the component will automatically recognize available fuels in the incoming mass flows. Still to do:

- [x] update documentation.
- [x] create an example with a mixture of different fuels.

Resolve #77 
Resolve #76

### DIFF
--- a/doc/getting_started/tutorial_combustion_chamber.rst
+++ b/doc/getting_started/tutorial_combustion_chamber.rst
@@ -52,14 +52,14 @@ As components there are two sources required, one for the fresh air, one for the
 
 	nw.add_conns(sf_comb, amb_comb, comb_fg)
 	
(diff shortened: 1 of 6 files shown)


fwitte:
Resolve #79

Issue #88, opened 2019-10-20
Remove access barriers for new devs in several modules


fwitte:
Some modules have grown very much (components.py in special, nearly 11.000 lines). This may obstruct new devs in gaining a good overview on the code. The code should be reorganised in a more friendly way. Also, a template for custom components could be added.

fwitte:
Here is a draft, how the module should look like, e. g. as a basis for discussion at the upcoming dev-meeting in Berlin.
- components
  - turbomachinery.py
  - heat_exchanger.py
  - combustion.py
  - node.py
  - piping.py
  - subsystems.py
  - subsystem_interface.py
- connections
  - bus.py
  - connection.py
- tools
  - fluid_property.py
  - data_container.py
  - helpers.py
  - logger.py
  - characteristics.py
  - data.??? (e. g. containing predefined characteristic lines)
- networks
  - network.py
  - network_reader.py
  - others?

fwitte:
Updated the component list  @MarBrandt.

fwitte:
The restructuring has been implemented on the dev_experimental branch and is ready for discussion. The docs review is issued in #87.

PR #98, opened 2019-11-07
Feature/components remodelling

MarBrandt:
splitting the components module in small units (see #88)


fwitte:
As the remodeling requires many changes, we will merge all PRs into dev_experimental first. We can then fix remaining issues before merging into dev.

Issue #87, opened 2019-10-20
Full documentation review


fwitte:
A full documentation review should be issued before the release of the next major version of TESPy. We will start the review the upcoming [dev-meeting](https://github.com/oemof/organisation/wiki/Meeting-December-2019) in December. Specific TODOs are listed below. I have added a short task description, feel free to comment/ask question. More specific information will be given at the dev-meeting.

**workflow**

0. If you did not fork oemof/tespy before:
   - fork tespy to your repositories
1. clone your repository
   - https: `git clone https://github.com/YOURUSERNAME/tespy.git` or use
   - ssh: `git@github.com:YOURUSERNAME/tespy.git`
2. if the branch oemof_doc_review is in your fork:
    - check out your oemof_doc_review branch (`git checkout -b oemof_doc_review origin/oemof_doc_review`)
2. Else:
    - create new local branch and pull from upstream (to add before).
    - `git remote add upstream https://github.com/oemof/tespy.git`
    - `git checkout -b oemof_doc_review`
    - `git pull upstream/oemof_doc_review`
3. push your adjustments to your remote branch (`git push` or `git push --set-upstream origin oemof_doc_review`)
4. create a PR on oemof/tespy to merge your remote branch in oemof/tespy oemof_doc_review
5. profit :)! (see https://github.com/oemof/oemof/pull/661)

**documentation review**

**online documentation**
In the online documentation we need to check for

- typos,
- comprehensibility (would a new user understand the main ideas?) and
- structure (does the structure of the chapters help the user understand how to use tespy?).

- [x] README/Introduction
- [x] Examples
- [x] Tutorials
- Installation and setup
  - [x] Revision
  - [ ] Add missing part for Mac

- Using TESPy
  - [x] Introduction
  - [x] networs and network reader
  - [x] components
  - [x] subsystems
  - [x] connections
  - [x] fluid properties

- What's New
  - [ ]  add examples regarding API changes

**doc-strings and inline comments**
In the doc-strings and the inline comments we need to check for the following:
- Are there any typos?
- Are the examples comprehensible?
- Is information well placed (are the equations and parameters apparent)?
- Are the links in the doc-strings working?
- Are the inline comments clear?

- tespy/components/
  - [ ] components.py
  - [ ] turbomachinery.py
  - [ ] heat_exchangers.py
  - [ ] piping.py
  - [ ] combustion.py
  - [ ] reactors.py
  - [ ] nodes.py
  - [ ] subsystem_interfaces.py
[...]

PR #112, opened 2019-12-04
Review the online documentation of tespy

fwitte:
This PR is related to #87.


fwitte:
The doc_review session brought up, that the table of contents of the online documentation is not very clear for new users. My proposal for the new table of contents looks like this (1-2 and 6-8 stay the same):

1. TESPy - Thermal Engineering Systems in Python
2. Installation
3. First steps using TESPy (contains a simple example for your first TESPy script)
4. Tutorial and Models/Examples (tutorials first, then examples)
5. TESPy (sub-)packages (detailed description of functionalities)
6. Developing TESPy
7. What's New
8. API

FranziPl:
I like it, I think it is intuitive.

jakob-wo:
I like the rearranged menu. 
Additionally I suggest to rename the first (1.) menu item "About" or "About TESPy". 
Among its sub-menu-items I would remove "Installation", "Documentation" and "Examples" because I find it confusing to find these items two times. We have them at the first level of the menu structure  already or in case of "Documentation" I find it confusing having a menu item called that way when I am reading the docu already - does that make sense? :)

fwitte:
Thank you both for your feedback.
> Additionally I suggest to rename the first (1.) menu item "About" or "About TESPy".
Good one, thank you :).
> Among its sub-menu-items I would remove "Installation", "Documentation" and "Examples" because I find it confusing to find these items two times. We have them at the first level of the menu structure already or in case of "Documentation" I find it confusing having a menu item called that way when I am reading the docu already - does that make sense? :)
This is due to the landingpage being (more or less) the same as the github README. I will adapt this.

fwitte:
A short recap for all TODOs:

Review
- [ ] Introduction
- [ ] Installation
- [ ] Connections

Update
- [ ] Components: characteristics
- [ ] Components: custom components
- [ ] Subsystems description (#124)
- [ ] What's New: Complete documentation of changes between 0.1.x and 0.2.x
- [ ] Others: Add an example for a coupled TESPy and solph model

Add
- [ ] Characteristics description

PR #122, opened 2019-12-10
Add a data module for tespy

fwitte:
The data module will contain default sets for component and bus characteristics. Additionally, standard diameters, pipe roughness, insulation material values etc. can be placed there.


fwitte:
This PR will be worked on together with #121.

PR #125, opened 2019-12-10
Fix imports in connections.rst

jfreissmann:


doc/using_tespy/connections.rst
```diff
@@ -18,28 +18,31 @@ As mentioned in the introduction, for each connection you can specify the follow
...
-and dc_flu for fluid composition. You need to import the :code:`hlp` module, if you want to specify data_containers.
+and dc_flu for fluid composition. If you want to specify data_containers, you need to import them from the :code:`tespy.tools` module.
 
 .. code-block:: python
 
+	from tespy.tools import dc_prop
+	from tespy.connections import ref
+
 	# set pressure and vapour mass fraction by value, temperature and enthalpy analogously
 	myconn.set_attr(p=7, x=0.5)
```
fwitte: I think it would be best to actually create a connection given two components, so if you copy the code into your editor, the python script will not raise errors.

doc/using_tespy/connections.rst
```diff
@@ -18,28 +18,31 @@ As mentioned in the introduction, for each connection you can specify the follow
...
-	myconn.set_attr(p=hlp.dc_prop(val=7, val_set=True), x=hlp.dc_prop(val=0.5, val_set=True))
-	myconn.set_attr(m=hlp.dc_prop(val0=10), p=hlp.dc_prop(val0=15), h=hlp.dc_prop(val0=100))
+	myconn.set_attr(p=dc_prop(val=7, val_set=True), x=dc_prop(val=0.5, val_set=True))
+	myconn.set_attr(m=dc_prop(val0=10), p=dc_prop(val0=15), h=dc_prop(val0=100))
 
 	# specify a value in a different unit for a specific parameter
-	myconn.set_attr(p=hlp.dc_prop(val=7, val_set=True, unit='MPa', unit_set=True)
+	myconn.set_attr(p=dc_prop(val=7, val_set=True, unit='MPa', unit_set=True)
 
 	# specify a referenced value: pressure of myconn is 1.2 times pressure at myotherconn minus 5 Pa (always SI unit here)
```
fwitte: Same here: Create the myotherconn, too.

doc/using_tespy/connections.rst
jfreissmann: Wouldn't you need to create the components as well? These examples could become to complex for new users.

fwitte: Well, at no part have I written, that you have to create the components first... If we add that in the introduction to this part, it would be pretty clear, I think.

fwitte:
Merged.

PR #146, opened 2020-01-31
Remove deprecated fields for characteristics data containers

maltefritz:
No warning/error is issued, when specifying x, y or x. y, z1 and z2 values for the datacontainers although these data are not used in any way.


fwitte:
The tests of the network reader are failing due to a bug in pandas to_csv() function (release 1.0.0), see https://github.com/pandas-dev/pandas/issues/31447. For now, I will adjust the version requirement for pandas to be below 1.0.0.

PR #147, opened 2020-02-02
Improve handling of component equations and derivatives

fwitte:
This PR changes the back end architecture for handing the equations and derivatives calls of the components. Additionally. the evaluation of equations and of the corresponding derivatives is modified: Equations regarding already converged variables will not be evaluated any more (or only every xth iteration), in order to cut calculation time. Therefore, I will add a performance test comparing the dev with this branch.

### DIFF
--- a/doc/tespy_modules/components.rst
+++ b/doc/tespy_modules/components.rst
@@ -133,25 +133,25 @@ need to specify the diameter the following way.
 
     # make diameter variable of system
     my_pipe.set_attr(pr=0.98, L=100, ks=0.00002, D='var')
-    
+
     # a second way of specifying this is similar to the
     # way used in the component parameters section
     # the benefit is, that val will be the starting value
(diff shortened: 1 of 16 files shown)


fwitte:
The component equations and derivatives have been remodeled. Time intensive calculation of equations and/or derivatives is skipped:
- equations: If the the equations residual is below a threshold and if the iteration count is not a multiple of 5.
- derivatives: If the change of a variable is near to zero the derivative of all functions to that variable is not updated anymore.

Next step is to modify the component node and the connection equations and derivatives analogously.

fwitte:
Still todo:

- [ ] update class node.
- [ ] update the online documentation of the tespy modules.

Performance-wise the PR looks very promising, especially if fluid mixtures and heat exchangers as well as turbomachinery is involved. The very large district heating network example (dhs from the oemof_examples repository) does not show much improvement as more iterations are required to find the solution.

time_comparison

PR #148, opened 2020-02-04
Add a new component: Evaporator in the ORC

ChaofanChen:
The new component of the evaporator has 3 inlets and 3 outlets. 2 inlets and 2 outlets are located in the hot side and 1 for steam 1 for liquid as the heat source. 1 inlet and 1 outlet are in the cold side, which is being evaporated.

In the steam outlet of the hot side 1, the default state is saturated liquid. In the outlet of the cold side, the default state is saturated gas.


fwitte:
Hi @ChaofanChen,
thank you for your contribution. We will need to apply the following changes, in order to merge the PR.

- As #147 looks very promising at the moment, the handling of derivatives and equations will need to be adapted to the back end changes applied in that PR.
- The class documentation and a minimal example are missing.
- The ttd_l, ttd_u and kA functionalities might be obsolete (as discussed).
- As this component is a somewhat unusual, I moved the code in a separate submodule (customs.py).

Have a nice day!

ChaofanChen:
Hi, @fwitte,
I think the 2 constrains are necessary. Specifying x is much more flexible, but x can not be equal to other values except 0 and 1 in outlet 1 and outlet 3. 

This is because, 1) for the condensate geo-steam in outlet 1, usually we also need to mix this part of the condensate geo-fluid with brine from the outlet2 after pressure rise through a pump. If the fluid out from the outlet1 is not fully condensed, then it will destroy the pressure-rise pump with 2-phase fluid in the pump. So x in outlet1 can only be 0.
2) For the outlet3 part, this component is evaporator, which means the default state of the working fluid should be just fully saturated gas state. If the users want to give the super heated gas state in the outlet3, then they can just give a Td_bp by setting the overheating is True. Meanwhile, the turbine also requests that the working fluid entering it should be fully gas state, 2-phase working fluid is not allowed to enter the turbine. So x in outlet 3 can only be 1.

ChaofanChen:
Hi @fwitte,

could you please have a look on errors of the test, I have no idea how to solve this problem. 

Thanks.

ChaofanChen:
Hi @fwitte ,

the derivatives and equations of orc_evaporator are modified according to the latest dev branch and has already rebased the 'dev' branch. 

Currently,  this PR is ready to merge from my side except some Coverage issues. Could you please have a look? 

Thanks.

fwitte:
Hi Chaofan,
I'll just update the todo-list.

- [x] As #147 looks very promising at the moment, the handling of derivatives and equations will need to be adapted to the back end changes applied in that PR.
- [x] The class documentation and a minimal example are missing.
- [ ] The evaporator image should be updated to look similar to the heat_exchanger component for conformity.
- [x] The ttd_l, ttd_u and kA functionalities might be obsolete (as discussed).
- [ ] Tests should be added to test the desired behavior of all equations.
- [x]<del>In order to add easier flexibility, we could think about removing the constraints regarding steam (saturated liquid state) and orc fluid (saturated gas state) outlet and instead let the user specify `x` at the corresponding connection. This would also remove the necessity of testing the parameters `subcooling` and `overheating` as the functionality is implemented in the connections already.</del>

ChaofanChen:
Hi @fwitte,

could you please have a look on the test of orc evaporator? 

I just want to test the vapour mass fraction in outlet 1 and outlets and also test the heat transfer. I am a little messy on the tests.

Thanks!

ChaofanChen:
Hi @fwitte,

I have already added tests related to the new evaporator component. I removed the kA functions in the orc evaporator as discussed before, you were right about that. Please have a look on this PR when you are available.

Thanks.

tespy/components/customs.py
```diff
@@ -0,0 +1,818 @@
...
+        **additional equations**
+
+        - :func:`tespy.components.customs.orc_evaporator.additional_equations`
+
+    Inlets/Outlets
+
+        - in1, in2, in3 (index 1: hot side 1, index 2: hot side 2,
+        index 3: cold side)
+        - out1, out2, out3 (index 1: hot side 1, index 2: hot side 2,
+        index 3: cold side)
```
fwitte: I think, we can additionally label these with working fluid and geothermal heat source.

tespy/components/customs.py
```diff
@@ -0,0 +1,818 @@
...
+
+    overheating : bool
+        Enable/disable overheating at oulet of the cold side,
+        default value: disabled.
+
+    Note
+    ----
+    The ORC evaporator are countercurrent heat exchangers.
+    Equation kA do not work for directcurrent and crosscurrent
+    or combinations of different types.
```
fwitte: Here we should add a more detailed description to the functionality of this component regarding partload performance. How is the evaporator built to ensure saturated gas state of the working fluid and saturated liquid state of the geothermal steam?

tespy/components/customs.py
```diff
@@ -0,0 +1,818 @@
...
+
+    Example
+    -------
+    A 2-phase geo-fluid is used as the heat source for evaporating
+    the working fluid. The evaporator is designed for calculate the
+    mass flow rate of the working fluid with known steam and
+    brine mass flow rate. From this, it is possible to calculate
+    the mass flow rate of the working fluid that is fully evaporated
+    through the ORC evaporator and its heat transfer coefficient.
```
fwitte: Heat transfer coefficient does not work at the moment. Maybe there is a way to calculate that value from measured data and implement an additional equation?

tespy/components/customs.py
```diff
@@ -0,0 +1,818 @@
...
+            0 = p_{3,in} \cdot pr3 - p_{3,out}
+
+        - :func:`tespy.components.customs.orc_evaporator.zeta_func`
+        - :func:`tespy.components.customs.orc_evaporator.zeta2_func`
+        - :func:`tespy.components.customs.orc_evaporator.zeta3_func`
+
+        **additional equations**
+
+        - :func:`tespy.components.customs.orc_evaporator.additional_equations`
+
```
fwitte: We can skip the usage of the additional equations. The additional equations are intended for components, that inherit some equations from a parent component but bring individual properties. For example: The fluid's state at the hot side of a heat_exchanger is not predefined. In a condenser however, the fluid is in saturated liquid state at that point. Therefore the condenser brings one additional equation to the base functionality of the heat_exchanger.

tespy/components/customs.py
```diff
@@ -0,0 +1,818 @@
...
+            o1 = self.outl[0].to_flow()
+            self.vec_res[k] = o1[2] - h_mix_pQ(o1, 0)
+            k += 1
+
+        ######################################################################
+        # equation for saturated gas at cold side outlet
+        if self.overheating.val is False:
+            o3 = self.outl[2].to_flow()
+            self.vec_res[k] = o3[2] - h_mix_pQ(o3, 1)
+            k += 1
```
fwitte: See comment on lines 61-65, we can move this into the equations part.

tespy/components/customs.py
```diff
@@ -0,0 +1,818 @@
...
+            self.mat_deriv[k, 3, 2] = 1
+            k += 1
+
+        ######################################################################
+        # derivatives for saturated gas at cold side outlet 3 equation
+        if self.overheating.val is False:
+            o3 = self.outl[2].to_flow()
+            self.mat_deriv[k, 5, 1] = -dh_mix_dpQ(o3, 1)
+            self.mat_deriv[k, 5, 2] = 1
+            k += 1
```
fwitte: See comment on lines 61-65, we can move this into the derivatives part.

tespy/components/customs.py
```diff
@@ -0,0 +1,818 @@
...
+            deriv[i + j, 4, j + 3] = -1
+            j += 1
+        # cold side
+        k = 0
+        for fluid in self.nw_fluids:
+            deriv[i + j + k, 2, k + 3] = 1
+            deriv[i + j + k, 5, k + 3] = -1
+            k += 1
+        return deriv.tolist()
+
```
fwitte: This is quite long. The is True for the implementation of the heat_exchanger class as well. We could change the base functionality of the fluid_deriv in the component class in order to work with all components that do not change fluid properties. The implementation of the fluid_deriv in the subsystem_interface class looks good for this purpose.
```
    def fluid_deriv(self):
        r"""
        Calculate the partial derivatives for all fluid balance equations.
        Returns
        -------
        deriv : list
            Matrix with partial derivatives for the fluid equations.
        """
        deriv = np.zeros((self.num_nw_fluids * self.num_i,
                          2 * self.num_i,
                          self.num_nw_vars))
        for i in range(self.num_i):
            for j in range(self.num_nw_fluids):
                deriv[i * self.num_nw_fluids + j, i, j + 3] = 1
                deriv[i * self.num_nw_fluids + j, self.num_i + i, j + 3] = -1
        return deriv.tolist()
```

tespy/components/customs.py
```diff
@@ -0,0 +1,818 @@
...
+        deriv : list
+            Matrix with partial derivatives for the mass flow balance
+            equations.
+        """
+        deriv = np.zeros((3, 6 + self.num_vars, self.num_nw_vars))
+        for i in range(self.num_i):
+            deriv[i, i, 0] = 1
+        for j in range(self.num_o):
+            deriv[j, j + i + 1, 0] = -1
+        return deriv.tolist()
```
fwitte: The same as above (sh)(c)ould be applied here.

tespy/components/customs.py
```diff
@@ -0,0 +1,818 @@
...
+        deriv = np.zeros((1, 6, self.num_nw_vars))
+        deriv[0, 2, 0] = self.numeric_deriv(self.bus_func, 'm', 2, bus=bus)
+        deriv[0, 2, 2] = self.numeric_deriv(self.bus_func, 'h', 2, bus=bus)
+        deriv[0, 5, 2] = self.numeric_deriv(self.bus_func, 'h', 5, bus=bus)
+        return deriv
+
+    def convergence_check(self, nw):
+        r"""
+        Performs a convergence check.
+        """
```
fwitte: This method is obsolete at the moment. A useful implementation would be changing the enthalpy value of outlet 3 to saturated gas state and the enthalpy value of the geothermal steam outlet to saturated liquid state.

tests/component_tests/orc_evaporator_tests.py
```diff
@@ -0,0 +1,129 @@
...
+               'hot side 2 (brine) '
+               'must be ' + str(round(zeta2, 1)) + ', is ' +
+               str(round(instance.zeta2.val, 1)) + '.')
+        eq_(round(instance.zeta2.val, 1), round(zeta2, 1), msg)
+        msg = ('Geometry independent friction coefficient at cold side '
+               '(Isopentane) must be ' + str(round(zeta3, 1)) + ', is ' +
+               str(round(instance.zeta3.val, 1)) + '.')
+        eq_(round(instance.zeta3.val, 1), round(zeta3, 1), msg)
+
+        shutil.rmtree('./tmp', ignore_errors=True)
```
fwitte: There should be tests regarding the subcooling and overheating parameters.

tespy/components/components.py
```diff
@@ -737,3 +737,44 @@ def zeta2_func(self):
...
+
+            .. math::
+
+                val = \begin{cases}
+                p_{in} - p_{out} & |\dot{m}| < \epsilon \\
+                \frac{\zeta_3}{D^4} - \frac{(p_{3,in} - p_{3,out}) \cdot \pi^2}
+                {8 \cdot \dot{m}_{3,in} \cdot |\dot{m}_{3,in}| \cdot
+                \frac{v_{3,in} + v_{3,out}}{2}} &
+                |\dot{m}| > \epsilon
+                \end{cases}
```
fwitte: This function is orc_evaporator specific. It should be either moved to that component or we must come up with a way, that the zeta_func handles all (zeta, zeta1, zeta2 and zeta3) function calls and apply this to the heat exchanger classes, too.

tespy/components/customs.py
ChaofanChen: Sure, that would be much more clear for the ORC cycle.

tespy/components/customs.py
ChaofanChen: Yes, I will add corresponding description according to the condenser.

tespy/components/customs.py
ChaofanChen: Currently, the kA value cannot be calculated because the principle of the ORC evaporator is still not clear for me. But it will be added after getting enough design parameters on the special component.

tespy/components/customs.py
ChaofanChen: Thanks for the explanation. Deleted.

tespy/components/customs.py
ChaofanChen: Moved to the equation part.

tespy/components/customs.py
ChaofanChen: Yes, move into derivatives part.

tespy/components/customs.py
ChaofanChen: Yes, thanks for that.

tespy/components/customs.py
ChaofanChen: Changed.

tespy/components/customs.py
ChaofanChen: Yes...., for that part, we still need to think about it.

tests/component_tests/orc_evaporator_tests.py
ChaofanChen: Added.

tespy/components/components.py
ChaofanChen: Yes, the specific zeta3 function for the orc_evaporator has been moved to the orc_evaporator class.

ChaofanChen:
> Additionally, there some more PEP issues (ignored by pep8speaks, I will look, if I can implement the missing issues) according to https://pep257.readthedocs.io/en/latest/error_codes.html. I will adjust the respective parts myself, as my editor is configured to display them anyways.

Thanks for the PEP issues in advance.

tests/component_tests/orc_evaporator_tests.py
```diff
@@ -0,0 +1,141 @@
...
+        eq_(round(instance.zeta2.val, 1), round(zeta2, 1), msg)
+        msg = ('Geometry independent friction coefficient at cold side '
+               '(Isopentane) must be ' + str(round(zeta3, 1)) + ', is ' +
+               str(round(instance.zeta3.val, 1)) + '.')
+        eq_(round(instance.zeta3.val, 1), round(zeta3, 1), msg)
+
+        # test parameters of 'subcooling' and 'overheating'
+        instance.set_attr(subcooling='Ture', overheating='Ture')
+        self.c2.set_attr(T=143)
+        self.c6.set_attr(T=119.7, state='g')
```
fwitte: Specifying the temperature difference to boiling point is more intuitive here. If the pressure changes, the specified temperature might be liquid fluid. The keyword `state='g'` is not required anymore in that case.

tests/component_tests/orc_evaporator_tests.py
```diff
@@ -0,0 +1,141 @@
...
+               str(round(instance.zeta2.val, 1)) + '.')
+        eq_(round(instance.zeta2.val, 1), round(zeta2, 1), msg)
+        msg = ('Geometry independent friction coefficient at cold side '
+               '(Isopentane) must be ' + str(round(zeta3, 1)) + ', is ' +
+               str(round(instance.zeta3.val, 1)) + '.')
+        eq_(round(instance.zeta3.val, 1), round(zeta3, 1), msg)
+
+        # test parameters of 'subcooling' and 'overheating'
+        instance.set_attr(subcooling='Ture', overheating='Ture')
+        self.c2.set_attr(T=143)
```
fwitte: Same here, temperature difference to boiling point should be negative.

tests/component_tests/orc_evaporator_tests.py
ChaofanChen: Yes, changed.

tests/component_tests/orc_evaporator_tests.py
ChaofanChen: Changed. But I'm a little confused why the negative value of subcooling should be lower than -3 or smaller ones? If I set 0.0 for subcooling, then it cannot be solved.

fwitte: 0.0 should be impossible as that would yield boiling point temperature. I will look into this.

fwitte: The issue derives from the specification of the bus value with the heat transfer specified at the same time. The network can not be solved in that state and so are all following calculations. I will fix this and push my commits.

PR #157, opened 2020-02-25
Add new feature: Bus efficiency printout

fwitte:
This PR implements an efficiency printout for busses. The networks.print_results() method and the bus_func() methods of components are adjusted.


fwitte:
Resolve #157

PR #159, opened 2020-02-26
Add new feature: extrapolation for characteristics

maltefritz:
**General**

Describe your pull request as transparent as possible:

* What functionality does it implement? It implements extrapolation of characteristic lines outside of the bounds of the upper and lower limits
* Where is it located? It is located in characteristics.py
* How does the API look? Unsure
* Which bugs are fixed? False results for characteristic lines outside of the bounds
* Are there related issues? Unsure

**Documentation**

Please document all changes and/or new features:

* Update the API documentation - Not yet
* Update the corresponding sections in the online documentation - Not yet
* Add inline comments to your code, if required - Not yet
* Update the corresponding What's New file - Not yet

**Testing**

* Adapt the software tests (including doc-tests), if the PR does affect existing tests - Not yet
* Add new tests for new features - Not yet

**Questions and Comments**

You are welcome to ask for help or start a discussion on a feature. We very much appreciate your contribution.

### DIFF
--- a/doc/tespy_modules/characteristics.rst
+++ b/doc/tespy_modules/characteristics.rst
@@ -19,16 +19,19 @@ the corresponding y-values.
 .. math::
 
     y = y_0 + \frac{x-x_0}{x_1-x_0} \cdot \left(y_1-y_0 \right)
-	
-If the x value is above the maximum x value or below the minium x value of the
-characteristic line the y value corresponding to the the maximum/minimum value
-is returned instead.
+
+It is possible to specify an :code:`extrapolate` parameter. If the value is
+:code:`False` (default state) and the x-value is above the maximum or below the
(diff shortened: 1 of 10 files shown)


fwitte:
Thank you for opening the PR. This is a good idea, I think. We can implement the changes in version 0.2.2. As long as the default value of extrapolate is False, it should not affect the results of models created with prior versions. I would like to add the feature to the characteristic maps, too.

Also, extrapolation to negative values could be an issue here. We might need to add a check, if a value is smaller than 0.

Please make the following updates to your PR:

- [ ] class documentation with extrapolate parameter description.
- add some tests in the characteristics_tests.py

  - [ ] testing lower and upper limits with and without extrapolation activated.

- [ ] update the v0.2.2.rst What's New file with an example how to use the extrapolation feature.

Thank you!

PR #174, opened 2020-04-07
New feature: CoolProp backend selection

### LINKED ISSUE #170: The specification of the CoolProp backend can save calculation time
The user should be able to chose the CoolProp backend manually. The options would be:

- `HEOS::fluid` (default, full EOS)
- `TTSE::fluid` (tabular data interpolation)
- `IF97::water` (only water is available for this case).

The default method would be HEOS (as it is at the moment). Using `TTSE::` or `IF97::` back-end, calculation times could be cut significantly.

fwitte:
The changes in this PR allow to select from the different backends in CoolProp and resolves #170.


fwitte:
Additional tests are needed to check
- performance differences in the backends and
- result accuracy.

fwitte:
Furthermore a test should be added, that tests if the mixture of a tespy_fluid with another fluid yields the same results as a mixture of the single fluids in regards to the method s_mix_pT.

PR #179, opened 2020-04-17
Architecture: Update package skeleton

fwitte:
This PR updates the package skeleton for consistency with the other oemof packages, e. g. https://github.com/oemof/oemof-solph.


fwitte:
The rtd build reads additional requirements from the rdt_requirements file. Maybe it should point to docs/requirements.txt

PR #205, opened 2020-06-02
Highlight result table information

jbueck:
Modifies the result table creation in networks.py using minor changes and colors set values blue and variables red depending on whether is_set or is_var are true. Uses bcolors package.

### DIFF
--- a/docs/whats_new/v0-3-2.rst
+++ b/docs/whats_new/v0-3-2.rst
@@ -22,6 +22,9 @@ New features
         coloring['err'] = '\033[0m'
         coloring['var'] = '\033[0m'
 
+  (See `here <https://en.wikipedia.org/wiki/ANSI_escape_code#Colors>`_ for a list of
+  available codes)
+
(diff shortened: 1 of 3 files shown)


fwitte:
Thanks @jbueck, this looks like a great addition. It seems as if there are missing dependencies in the setup.py. I'll have a closer look at this next week as I am on holiday this week :). Also, maybe we can make the color flexible to the user's choice (keeping red and blue as default).
Best regards
Francesco

fwitte:
Hi @jbueck,
thank you again, very nice feature. I adjusted the order of imports and modified some lines to always print floats. In some instances the results table had integers and floats mixed which messed up the alignment for me. I also update the What's New section.

Do you happen to have a link which we can provide to the users, where to lookup the available color codes? Is it only 30-37 or are there more? If you like, just add the link at the end of the text on the new feature. Also, if you want your full name in the contributors list, just update your entry.
Thank you again and have a nice week, best regards
Francesco

fwitte:
Thank you very much for your contribution, merged into dev!

PR #215, opened 2020-08-06
Add new feature: exergy analysis

juliusmeier:
Add methods and tools for exergy analysis of processes.


fwitte:
Thank you very much for the suggested features. I'll make some annotations to the code you supplied, so we can merge your feature into the code. Also, we will need to define tests for checking actual vs. expected results.

src/tespy/components/heat_exchangers.py
```diff
@@ -867,6 +867,26 @@ def calc_parameters(self):
 
         self.check_parameter_bounds()
 
+    def exergy_balance(self, Tamb):
+        r"""
+        Perform exergy balance of a simple heat exchanger.
+
```
fwitte, line 873: I think "Calculate" suits better here.

src/tespy/components/heat_exchangers.py
```diff
@@ -867,6 +867,26 @@ def calc_parameters(self):
...
+            .. math::
+
+                T_m = \frac{T_{ein} - T_{aus}}{\ln\left({\frac{T_{ein}}
+                                                         {T_{aus}}}\right)}
+                Ex_{output} = \dot{m}_{in} \cdot (ex_{ph,out} - ex_{ph,in})
+                Ex_{input} = \left(1 - \frac{T_{amb}}
+                                   {T_m}\right) \cdot \dot{Q}
+                Ex_{loss} = 0
```
fwitte, line 881: These lines do not get linebreaks in LaTeX, we need to add extra newlines or "\\" at the end of a line (see https://tespy--215.org.readthedocs.build/en/215/api/tespy.components.html#tespy.components.heat_exchangers.condenser.exergy_balance).

src/tespy/components/components.py
```diff
@@ -586,6 +586,12 @@ def initialise_fluids(self, nw):
     def convergence_check(self, nw):
         return
 
+    def exergy_balance(self):
```
fwitte, line 589: For consistency, all exergy balance methods should have the same set of parameters (`self` and `Tamb`).

src/tespy/components/heat_exchangers.py
```diff
@@ -867,6 +867,26 @@ def calc_parameters(self):
...
+        """
+        Tm = (self.inl[0].T.val_SI - self.outl[0].T.val_SI) / (
+            np.log(self.inl[0].T.val_SI / self.outl[0].T.val_SI))
+        self.Ex_output = self.inl[0].m.val_SI * (self.outl[0].ex_physical
+                                                 - self.inl[0].ex_physical)
```
fwitte, line 886: Having the linebreak after the first parenthesis would be more consistent with the other line, I think.

src/tespy/components/heat_exchangers.py
```diff
@@ -867,6 +867,26 @@ def calc_parameters(self):
...
+        self.Ex_input = (1 - (Tamb / Tm)) * self.Q.val
+        self.Ex_loss = 0
```
fwitte, line 888: The exergy loss is not necessarily 0, I think. What about e.g. pressure losses when no heat is transferred? My suggestions would be:

- Q > 0 -> `self.Ex_input = (1 - (Tamb / Tm)) * self.Q.val`, `self.Ex_output = self.inl[0].m.val_SI * (self.outl[0].ex_physical - self.inl[0].ex_physical)` and `self.Ex_loss = self.Ex_input - self.Ex_output`
- Q = 0 -> `self.Ex_input = 0`, `self.Ex_output = 0` and `self.Ex_loss = - self.inl[0].m.val_SI * (self.outl[0].ex_physical - self.inl[0].ex_physical)`
- Q < 0 -> `self.Ex_input = self.inl[0].m.val_SI * (self.outl[0].ex_physical - self.inl[0].ex_physical)`, `self.Ex_output = (1 - (Tamb / Tm)) * self.Q.val` and `self.Ex_loss = self.Ex_output - self.Ex_input`

What do you think?

src/tespy/components/heat_exchangers.py
```diff
@@ -867,6 +867,26 @@ def calc_parameters(self):
...
+
+                T_m = \frac{T_{ein} - T_{aus}}{\ln\left({\frac{T_{ein}}
+                                                         {T_{aus}}}\right)}
+                Ex_{output} = \dot{m}_{in} \cdot (ex_{ph,out} - ex_{ph,in})
+                Ex_{input} = \left(1 - \frac{T_{amb}}
+                                   {T_m}\right) \cdot \dot{Q}
+                Ex_{loss} = 0
+        """
+        Tm = (self.inl[0].T.val_SI - self.outl[0].T.val_SI) / (
+            np.log(self.inl[0].T.val_SI / self.outl[0].T.val_SI))
```
fwitte, line 884: I am not sure about this calculation. Given dQ = TdS, Tm would be Q / S or (h2 - h1) / (s2 - s1). But I am not sure in which way this is affected by entropy production by pressure losses and if the equation is even applicable in that case. Do you have any literature at hand? I'll try to see, what I can find, too.

src/tespy/components/heat_exchangers.py
```diff
@@ -1263,6 +1283,27 @@ def calc_parameters(self):
...
+        Tamb : float
+            Ambient temperature Tamb / K.
+
+            .. math::
+
+                Ex_{output} = \dot{m}_{in} \cdot (ex_{ph,out} - ex_{ph,in})
+                Ex_{input} = \dot{Q}_{sol} \cdot \left(1 - \frac{4}
+                                                       {3} \frac{T_{amb}}
+                                                       {T_{sun}}\right)
+                Ex_{loss} = 0
```
fwitte, line 1299: Same here with the linebreaks, also can you add the source for this equation in the .bib file and a citation here?

src/tespy/components/heat_exchangers.py
```diff
@@ -1263,6 +1283,27 @@ def calc_parameters(self):
...
+        """
+        T_sun = 5679
+        self.Ex_output = self.inl[0].m.val_SI * (self.outl[0].ex_physical
+                                                 - self.inl[0].ex_physical)
+        self.Ex_input = self.Q.val * (1 - ((4/3) * (Tamb/T_sun)))
+        self.Ex_loss = 0
```
fwitte, line 1305: - linebreak after parenthesis
- missing spaces around 4 / 3 and Tamb / Tsun
- similar issue with the exergy balance here. Although Q should only be > 0, there might be exergy losses due to pressure losses we do not yet consider.

src/tespy/components/heat_exchangers.py
```diff
@@ -2477,6 +2518,24 @@ def calc_parameters(self):
 
         self.check_parameter_bounds()
 
+    def exergy_balance(self):
```
fwitte, line 2521: Needs `Tamb` as parameter, too.

src/tespy/components/heat_exchangers.py
```diff
@@ -2477,6 +2518,24 @@ def calc_parameters(self):
...
+        r"""
+        Perform exergy balance of a heat exchanger.
+
+            .. math::
+
+                Ex_{output} = \dot{m}_{in,cold} \cdot (ex_{ph,out,cold}
+                                                       - ex_{ph,in,cold})
+                Ex_{input} = \dot{m}_{in,hot} \cdot (ex_{ph,in,hot}
+                                                     - ex_{ph,out,hot})
+                Ex_{loss} = 0
```
fwitte, line 2531: Linebreaks

src/tespy/components/heat_exchangers.py
```diff
@@ -2885,6 +2944,21 @@ def ttd_u_func(self):
         T_o2 = T_mix_ph(o2, T0=self.outl[1].T.val_SI)
         return self.ttd_u.val - T_bp_p(i1) + T_o2
 
+    def exergy_balance(self):
```
fwitte, line 2947: Tamb

src/tespy/components/heat_exchangers.py
```diff
@@ -2885,6 +2944,21 @@ def ttd_u_func(self):
...
+        r"""
+        Perform exergy balance of a condenser.
+
+            .. math::
+
+                Ex_{output} = 0
+                Ex_{input} = \dot{m}_{in} \cdot (ex_{ph,in} - ex_{ph,out})
+                Ex_{loss} = 0
```
fwitte, line 2955: Linebreaks

src/tespy/components/heat_exchangers.py
```diff
@@ -2885,6 +2944,21 @@ def ttd_u_func(self):
...
+        """
+        self.Ex_output = 0
```
fwitte, line 2957: I do not really like `Ex_output = 0` here, as the condenser might be used for heat production in district heating. In that case, Ex is clearly not 0.

src/tespy/components/piping.py
```diff
@@ -534,3 +534,18 @@ def calc_parameters(self):
         self.Sirr.val = i[0] * (s_mix_ph(o) - s_mix_ph(i))
 
         self.check_parameter_bounds()
+
+    def exergy_balance(self):
```
fwitte, line 538: Tamb

src/tespy/components/piping.py
```diff
@@ -534,3 +534,18 @@ def calc_parameters(self):
...
+        r"""
+        Perform exergy balance of a valve.
+
+         .. math::
+
+                Ex_{output} = 0
+                Ex_{input} = \dot{m}_{in} \cdot (ex_{ph,in} - ex_{ph,out})
+                Ex_{loss} = 0
```
fwitte, line 546: Linebreaks

src/tespy/components/piping.py
```diff
@@ -534,3 +534,18 @@ def calc_parameters(self):
...
+        """
+        self.Ex_output = 0
+        self.Ex_input = self.inl[0].m.val_SI * (self.inl[0].ex_physical -
+                                                self.outl[0].ex_physical)
+        self.Ex_loss = 0
```
fwitte, line 551: Exergy losses should be `self.inl[0].m.val_SI * (self.inl[0].ex_physical - self.outl[0].ex_physical)`, input and output would be 0.

src/tespy/components/turbomachinery.py
```diff
@@ -1269,6 +1269,21 @@ def calc_parameters(self):
 
         self.check_parameter_bounds()
 
+    def exergy_balance(self):
```
fwitte, line 1272: Tamb

src/tespy/components/turbomachinery.py
```diff
@@ -1269,6 +1269,21 @@ def calc_parameters(self):
...
+        r"""
+        Perform exergy balance of a pump.
+
+            .. math::
+
+                Ex_{output} = \dot{m}_{in} \cdot (ex_{ph,out} - ex_{ph,in})
+                Ex_{input} = P
+                Ex_{loss} = 0
```
fwitte, line 1280: Linebreaks

src/tespy/components/turbomachinery.py
```diff
@@ -1724,3 +1739,23 @@ def calc_parameters(self):
             self.eta_s_char.func.get_bound_errors(expr, self.label)
 
         self.check_parameter_bounds()
+
+    def exergy_balance(self):
```
fwitte, line 1743: Tamb

src/tespy/components/turbomachinery.py
```diff
@@ -1724,3 +1739,23 @@ def calc_parameters(self):
...
+        r"""
+        Perform exergy balance of a turbine.
+
+            .. math::
+
+                Ex_{output} = P
+                Ex_{input} = \dot{m}_{in} \cdot (ex_{ph,in} - ex_{ph,out})
+                Ex_{loss} = 0
+
```
fwitte, line 1752: Linebreaks

src/tespy/components/turbomachinery.py
```diff
@@ -1269,6 +1269,21 @@ def calc_parameters(self):
...
+                Ex_{output} = \dot{m}_{in} \cdot (ex_{ph,out} - ex_{ph,in})
+                Ex_{input} = P
+                Ex_{loss} = 0
+        """
+        self.Ex_output = self.inl[0].m.val_SI * (self.outl[0].ex_physical
+                                                 - self.inl[0].ex_physical)
+        self.Ex_input = self.P.val
+        self.Ex_loss = 0
+
```
fwitte, line 1286: Ex_loss would be Ex_input - Ex_output for my understanding.

src/tespy/components/turbomachinery.py
```diff
@@ -1724,3 +1739,23 @@ def calc_parameters(self):
...
+        Note
+        ----
+        Exergy destruction is always positive. Therefore, the output power
+        value needs to be multiplied by -1.
+        """
+        self.Ex_output = self.P.val*-1
+        self.Ex_input = self.inl[0].m.val_SI * (self.inl[0].ex_physical
+                                                - self.outl[0].ex_physical)
+        self.Ex_loss = 0
```
fwitte, line 1761: Ex_loss would be Ex_input - Ex_output for my understanding.

src/tespy/connections.py
```diff
@@ -29,6 +29,7 @@
 from tespy.tools.data_containers import dc_prop
 from tespy.tools.data_containers import dc_simple
 from tespy.tools.helpers import TESPyConnectionError
+from tespy.tools.fluid_properties import calc_physical_exergy
```
fwitte, line 32: This line needs to be moved one up (between data_containers and helpers) as the imports should be sorted alphabetically.

src/tespy/tools/fluid_properties.py
```diff
@@ -1729,3 +1729,35 @@ def ds_mix_pdT(flow, T):
...
+        Specific physical exergy ex_ph / (J/kg).
+
+        .. math::
+
+            ex_{ph} = (h - h(p_{amb}, T_{amb})) -
+            T_{amb} \cdot (s - s(p_{amb}, T_{amb}))
+    """
+    hamb = h_mix_pT([0, pamb, 0, conn.fluid.val], Tamb)
+    samb = s_mix_pT([0, pamb, 0, conn.fluid.val], Tamb)
+    return round((conn.h.val_SI - hamb) - Tamb*(conn.s.val_SI - samb), 6)
```
fwitte, line 1763: - I am not sure, if we want to round here. I suggest to not round here, but at a later state if necessary.
- Missing spaces around `...Tamb * (...`

src/tespy/components/turbomachinery.py
```diff
@@ -1269,6 +1269,32 @@ def calc_parameters(self):
...
+                E_{P} = \dot{m}_{in} \cdot (e_{ph,out} - e_{ph,in})
+                E_{F} = P
+        """
+        self.E_P = self.inl[0].m.val_SI * (self.outl[0].ex_physical
+                                                 - self.inl[0].ex_physical)
+        if bus == None:
+            self.E_F = self.P.val
+        else:
+            self.E_F = self.calc_bus_value(bus)
+
```
fwitte, line 1297: This is, so instead of the pump power the exergy input can be the electrical power of the motor?

src/tespy/networks/networks.py
```diff
@@ -2511,6 +2511,134 @@ def postprocessing(self):
...
+        - Get values for physical exergy of connections.
+        - Call exergy balances of components.
+          The components cycle\_closer, source, sink and splitter are being
+          skipped because it does not make sense to perform an exergy analysis
+          with these components.
+        - Calculate exergy destruction and exergetic efficiency of components.
+          If there is a componenten in the network for which there is no
+          exergy balance implemented in TESPy, 0 is assigned to E_D, E_F and
+          E_P of the componenten. The exergetic efficiency epsilon gets the
+          value -111.
```
fwitte, line 2527: I am not so sure about this. Can we use np.nan instead?

src/tespy/networks/networks.py
```diff
@@ -2511,6 +2511,134 @@ def postprocessing(self):
...
+        for cp in self.comps.index:
+            if (isinstance(cp, cycle_closer) or isinstance(cp, sink) or
+                    isinstance(cp, source) or isinstance(cp, splitter)):
+                continue
+            cp.exergy_balance(Tamb, bus)
+
+            # calculate E_D for components
+            try:
+                if isinstance(cp.E_P, str):
+                    cp.E_D = cp.E_F
```
fwitte, line 2589: Why do we check for a string. Is it 'n/a'? Assigning np.nan instead and checking for isnan would be prettier, I think.

src/tespy/networks/networks.py
```diff
@@ -2511,6 +2511,134 @@ def postprocessing(self):
...
+                else:
+                    cp.E_D = cp.E_F - cp.E_P
+            except AttributeError:
+                cp.E_D = 0
+                cp.E_F = 0
+                cp.E_P = 0
```
fwitte, line 2595: I would like to remove the try except. The exergy_balance base method can assign these values for components that do not yet have the exergy balance implemented. As it is called prior to this line, there should not be an issue.

src/tespy/networks/networks.py
```diff
@@ -2511,6 +2511,134 @@ def postprocessing(self):
...
+            # calculate epsilon for components
+            try:
+                if isinstance(cp.E_P, str):
+                    cp.epsilon = 'n/a'
+                else:
+                    cp.epsilon = cp.E_P / cp.E_F
+            except ZeroDivisionError:
+                cp.epsilon = -111
```
fwitte, line 2603: Same here, I prefer using np.nan if possible.

src/tespy/networks/networks.py
```diff
@@ -2511,6 +2511,134 @@ def postprocessing(self):
...
+            if cp.epsilon == -111:
+                continue
+            self.E_D_sum += cp.E_D
+
+        # calculate E_P, E_F, E_L, E_D and epsilon for network
+        self.E_L = 0
+        self.E_F = 0
+
+        for b in self.busses.values():
+            self.E_P = b.P.val*-1
```
fwitte, line 2620: Some busses may not be used for exergy balancing. E.g. a bus can be used, to relate the power values of two components but is not connected to any physical out- or input of energy of the network. A solution could be, that the user needs to pass the busses, that should be considered in this method.

src/tespy/networks/networks.py
```diff
@@ -2511,6 +2511,134 @@ def postprocessing(self):
...
+
+        # calculate exergy destruction ratios for components
+        for cp in self.comps.index:
+            if (isinstance(cp, cycle_closer) or isinstance(cp, sink) or
+                    isinstance(cp, source) or isinstance(cp, splitter)):
+                continue
+            if cp.epsilon == -111:
+                continue
+            cp.y_Dk = cp.E_D / self.E_F
+            cp.ystar_Dk = cp.E_D / self.E_D
```
fwitte, line 2640: On top of the equations we need to have a detailed description of the parameters we calculate, so it is clear what each of the ratios and numbers mean.

src/tespy/networks/networks.py
```diff
@@ -2606,6 +2734,78 @@ def print_components(c, *args):
...
+    def print_exergy_conns(self):
+        r"""Print the calculation results of the (specific) physical exergy of
+        the connections to prompt.
+        """
+        df = pd.DataFrame(columns=['e_PH / (kJ / kg)', 'E_PH / MW'])
+        for c in self.conns.index:
+            row = (c.source.label + ':' + c.source_id + ' -> ' +
+                   c.target.label + ':' + c.target_id)
+            row_data = [c.ex_physical/10**3, c.Ex_Physical/10**6]
+            df.loc[row] = row_data
```
fwitte, line 2746: Flexible unit system would be nice. The user will provide the unit and the method adapts correspondingly. I'll set this up.

src/tespy/networks/networks.py
```diff
@@ -2606,6 +2734,78 @@ def print_components(c, *args):
...
+            if not isinstance(cp.E_P, str):
+                cp.E_P = round(cp.E_P / 10**6, 4)
+            if not isinstance(cp.epsilon, str):
+                cp.epsilon = round(cp.epsilon, 4)
+            if cp.E_D < E_D_min:
+                continue
+
+            row_data = [cp.label, cp.E_P, cp.E_F/10**6,
+                        cp.E_D/10**6, cp.epsilon, cp.y_Dk,
+                        cp.ystar_Dk]
```
fwitte, line 2787: I will move parts of this into the exergy_analysis, so we do not need to loop through the components so many times.

src/tespy/components/heat_exchangers.py
```diff
@@ -2885,6 +2942,34 @@ def ttd_u_func(self):
...
+        ----
+            .. math::
+
+                E_{P} = n/a
+                E_{F} = \dot{m}_{in} \cdot (e_{ph,in} - e_{ph,out})
+        """
+        if self.dissipative == True:
+            self.E_P = 'n/a'
+        if self.dissipative == False:
+            self.E_P = self.inl[1].m.val_SI * (self.outl[1].ex_physical
```
fwitte, line 2968: use if ... else ... Where is self.dissipative defined? And how to change it?

src/tespy/connections.py
```diff
@@ -596,6 +597,27 @@ def to_flow_design(self):
...
+            Ambient temperature Tamb / K.
+        
+        Note
+        ----
+            .. math::
+
+                E^{\text{PH}} = \dot{m} \cdot e^{\text{PH}}
+        """
+        self.ex_physical = calc_physical_exergy(self, pamb, Tamb)
+        self.Ex_Physical = self.m.val_SI * self.ex_physical
```
fwitte, line 620: This is fine, we do not need to move it absolute value to the calc_exergy method. Subsequently we can use something like self.inl[0].Ex_physical - self.outl[0].Ex_physical for the component calculations.

src/tespy/components/nodes.py
```diff
@@ -1415,6 +1415,37 @@ def additional_derivatives(self, increment_filter, k):
...
+            if self.inl[i].T.val_SI <= self.outl[0].T.val_SI:
+                self.E_P += self.inl[i].m.val_SI * (self.outl[0].ex_physical 
+                                                    - self.inl[i].ex_physical)
+
+        self.E_F = 0
+        for i in range(0, len(self.inl)):
+            if self.inl[i].T.val_SI > self.outl[0].T.val_SI:
+                self.E_F += self.inl[i].m.val_SI * (self.inl[i].ex_physical 
+                                                    - self.outl[0].ex_physical)
+
```
fwitte, line 1448: This might only work with pure fluids. When mixing different fuels the composition of the fluids at the inlets and the outlet will be different, and therefore the reference might not be correct anymore. We will need to check this.

src/tespy/components/turbomachinery.py
juliusmeier: Yes
If the energy flow is not in a bus, there is no electrical efficiency of the pump. The pump power is used for fuel exergy (=input exergy). If we have an electrical efficiency, the fuel exergy is the electrical power of the motor.

src/tespy/networks/networks.py
juliusmeier: Yes, of course.
I did not think of that approach. It convinces me.

src/tespy/networks/networks.py
juliusmeier: Good point. 
How can I access the net power of a bus? Is it the same way I did it here?

src/tespy/networks/networks.py
juliusmeier: Alright, I will describe it in more detail

src/tespy/networks/networks.py
juliusmeier: Thanks!

src/tespy/components/heat_exchangers.py
juliusmeier: Ah yes, if.. else.. is better.
self. dissipative is defined right above the method exergy_balance 
In the user's code it can be changed by condensername.dissipative = False
Then, the condenser is treated as an ordinary heat exchanger and not as a dissipative component.

src/tespy/connections.py
juliusmeier: Alright, I agree.

src/tespy/components/nodes.py
juliusmeier: Oh yes, might not work. I forgot the case of mixing different fluids

src/tespy/networks/networks.py
juliusmeier: Dissipative components such as condensers and valves do not have a poduct exergy (=exergy output) by definition. That is way I assign the string  "n/a"  for E_P in the exergy balance of these components. In the printed results, the user can see that it is intentional when using "n/a" instad of np.nan. 
For dissipatice components it is defined that all fuel exergy is destructed.

src/tespy/networks/networks.py
juliusmeier: Alright, I will think of a solution there.

src/tespy/networks/networks.py
juliusmeier: yes, I agree

src/tespy/components/turbomachinery.py
```diff
@@ -1269,6 +1269,30 @@ def calc_parameters(self):
...
+            E_{P} = \dot{m}_{in} \cdot \left(e_{ph,out} - e_{ph,in}\right)\\
+            E_{F} = P
+        """
+        self.E_P = self.inl[0].m.val_SI * (
+            self.outl[0].ex_physical - self.inl[0].ex_physical)
+        if bus is None:
+            self.E_F = self.P.val
+        else:
+            self.E_F = self.calc_bus_value(bus)
+
```
fwitte, line 1295: Implement identical balance for compressor

src/tespy/components/turbomachinery.py
```diff
@@ -1724,3 +1748,32 @@ def calc_parameters(self):
...
+            E_{P} = P\\
+            E_{F} = \dot{m}_{in} \cdot (e_{ph,in} - e_{ph,out})
+
+        Note
+        ----
+        Exergy destruction is always positive. Therefore, the output power
+        value needs to be multiplied by -1.
+        """
+        if bus == None:
```
fwitte, line 1774: is None

src/tespy/networks/networks.py
fwitte: We'll also need a literature reference for the notation.

src/tespy/networks/networks.py
```diff
@@ -2616,6 +2744,78 @@ def print_components(c, *args):
...
+
+        Parameters
+        ----------
+        E_D_min : float
+            Minimum exergy destruction to be printed to prompt.
+        sort_des : boolean
+            Sort the component results descending by exergy destruction.
+        """
+        df = pd.DataFrame(columns=['label', 'E_P / MW', 'E_F / MW',
+                                   'E_D / MW', 'epsilon', 'y_D,k', 'y*_D,k'])
```
fwitte, line 2781: Same with the units

src/tespy/networks/networks.py
```diff
@@ -2616,6 +2744,78 @@ def print_components(c, *args):
...
+            if sort_desc is True:
+                df.sort_values(by=['E_D / MW'], ascending=False, inplace=True)
+
+        print('\n##### RESULTS (components) Exergy analysis #####')
+        print(tabulate(df, headers='keys', tablefmt='psql', floatfmt='.4f',
+                       showindex=False))
+
+        # print network exergy analysis results
+        df = pd.DataFrame(columns=['label', 'E_P / MW', 'E_F / MW',
+                                   'E_D / MW', 'E_L / MW','epsilon'])
```
fwitte, line 2808: units

src/tespy/networks/networks.py
```diff
@@ -2616,6 +2744,78 @@ def print_components(c, *args):
...
+        row_data = ['network', self.E_P/10**6, self.E_F/10**6, self.E_D/10**6,
+                    self.E_L/10**6, self.epsilon]
+        df.loc['network'] = row_data
+        print('\n##### RESULTS (network) Exergy analysis #####')
+        print(tabulate(df, headers='keys', tablefmt='psql', floatfmt='.4f',
+                       showindex=False))
+        print(tabulate([['Bottom up calculation of exergy destruction / MW',
+                        self.E_D_sum/10**6 ]],tablefmt='psql', floatfmt='.4f',
```
fwitte, line 2816: units

src/tespy/tools/fluid_properties.py
```diff
@@ -1729,3 +1729,35 @@ def ds_mix_pdT(flow, T):
...
+
+    Returns
+    -------
+    e_ph : float
+        Specific physical exergy e_ph / (J/kg).
+
+        .. math::
+
+            e_{ph} = (h - h(p_{amb}, T_{amb})) -
+            T_{amb} \cdot (s - s(p_{amb}, T_{amb}))
```
fwitte, line 1759: \left( \right)

src/tespy/components/heat_exchangers.py
fwitte: As we do not necessarily know, which part of the change in entropy is from heat transfer and which part is from e.g. pressure losses, the method of just using mean temperature may be sufficient.

src/tespy/components/heat_exchangers.py
```diff
@@ -2942,33 +2933,32 @@ def ttd_u_func(self):
...
+            E_{F} = \dot{m}_{in} \cdot (e_{ph,in} - e_{ph,out})
         """
-        if self.dissipative == True:
-            self.E_P = 'n/a'
-        if self.dissipative == False:
-            self.E_P = self.inl[1].m.val_SI * (self.outl[1].ex_physical
-                                               - self.inl[1].ex_physical)
-        self.E_F = self.inl[0].m.val_SI * (self.inl[0].ex_physical
-                                           - self.outl[0].ex_physical)
+        if self.dissipative is True:
```
juliusmeier, line 2961: How can I set dissipative to False?

src/tespy/components/components.py
```diff
@@ -586,20 +586,20 @@ def initialise_fluids(self, nw):
...
         ----------
-        Tamb_val_SI : float
-            Ambient temperature in K.
         bus  : tespy.connections.bus
             Energy flows in network. Used to calculate product exergy
             or fuel exergy of turbines and pumps.
         """
-        logging.warning('Method not available for this component: '
-                        + self.label)
+        self.E_D = 0
```
juliusmeier, line 598: I think it is not necessary to define E_D here because it is calculated in the exergy_analysis method in networks.py anyway

juliusmeier:
Hello,
I edited a couple of things.

- For dissipative components I assigned n/d (not defined) for E_P and n/a (not available) for epsilon. That is how I would prefer it. Let me now if you dont like it or it doesnt fit with the unit conversion.
- I added the variable E_D_sum in the exergy analysis method. It vanished but the variable is important to sum up the E_D of components in order to compare it with the E_D_total from exergy balance of overall system.
- Regarding the merche it is necessary to use specific exergy e_ph in exergy balance of component. We might just use it everywhere else as well for consistency?
- I will revise the documentation when equations and everything else is finished. 
- The warning shows that there is no exergy balance for cycle_closer, sink, source and splitter which is correct but I am not sure if it is necessary.
- I would like to talk to you about busses, E_F with average Temperature approach, unit conversion and how to set the dissipative attribute.

src/tespy/components/nodes.py
```diff
@@ -1415,6 +1415,37 @@ def additional_derivatives(self, increment_filter, k):
...
+        for i in range(0, len(self.inl)):
+            if self.inl[i].T.val_SI <= self.outl[0].T.val_SI:
+                self.E_P += self.inl[i].m.val_SI * (self.outl[0].ex_physical
+                                                    - self.inl[i].ex_physical)
+
+        self.E_F = 0
+        for i in range(0, len(self.inl)):
+            if self.inl[i].T.val_SI > self.outl[0].T.val_SI:
+                self.E_F += self.inl[i].m.val_SI * (self.inl[i].ex_physical
+                                                    - self.outl[0].ex_physical)
```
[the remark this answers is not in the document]
juliusmeier: When mixing different fluids, we need to consider chemical exergy as well.
But only physical exergy is implemented. I would suggest to make a comment that the exergy analysis only works for pure fluids.

fwitte:
Hi @juliusmeier,
I tried contating you directly via mail but got an autoreply that your inbox is full and the mail could no be delivered. Can you write me a message, e.g. via telegram https://t.me/fwitte?
Best regards
Francesco

src/tespy/components/heat_exchangers.py
fwitte: At the moment it would be something like `yourcondenser.dissipative = False` after simulation of the network and prior to the exergy analysis. Would be better, if it is just specified by the `set_attr()` method like every other parameter. I'll take this on.

fwitte:
Hi @juliusmeier,
regarding the exergy analysis method, we need to make some additional considerations (already working on those), just wanted to let you know. Maybe you have some additional thoughts:
- Exergy fuel of the network can either be energy transfer to the network or a mass flow coming into the system (e.g. using compressed air from a storage tank)
- Exergy production of the network can either be energy transfer from the network or a mass flow leaving the network (e.g. loading compressed air into a storage tank)
- Exergy losses of the network are mass flows that leave the network into the ambient without use (e.g. flue gases of a gas turbine)
- Consider internal busses that neither represent exergy production nor fuel but still have exergy destruction (e.g. driving the feed water pump with a steam turbine on the same shaft and considering mechanical losses)

After implementation of these changes, we need one more check of the documentation for typos etc. and additionally I want to create three simple examples (maybe use the existing examples) to present the capabilities of the exergy analysis tool:
- Concentraing solarthermal power plant
- Heat pump or cooling unit
- Compressed air storage

Best
Francesco

fwitte:
>     * The warning shows that there is no exergy balance for cycle_closer, sink, source and splitter which is correct but I am not sure if it is necessary.

Good point, I do not like that either. We should either degrade that method to debug or just throw it out of the window.

juliusmeier:
Hello Francesco,
I have some comments.

- heat_exchanger: We need to add the case when a heat exchanger is used in refrigaration applications. Then it happens that the hot steam is cooled by the cold stream and all heat transfers occur at or below T0

-  sink: If a sink represents an exergy loss, it is regarded as E_D of the sink at the moment. That is not right. An exergy loss is only associated with the network. So, I would suggest that E_D = np.nan always and the exergy loss value is added to E_L of the network.

- merge: According to "Thermal design and optimization" by Bejan et al on page 152 it is also correct to say: E_P = E_out and E_F = sum(E_in)

- exergy balance of network: Right now E_D_network is the sum of E_D_comps. But we have to calculate the Network exergy balance separately. So E_D_network = E_F_network - E_P_network - E_L_network. Then we can print E_D_sum of components again as a value to compare with E_D_network.

fwitte:
Hi thanks for your comments,

>     * heat_exchanger: We need to add the case when a heat exchanger is used in refrigaration applications. Then it happens that the hot steam is cooled by the cold stream and all heat transfers occur at or below T0

We need to check this, thank you! For assessing heat transfer without knowlegde of both sides of the heat exchanger, this is a good read on the topic, I think: https://www.researchgate.net/publication/264441505_Understanding_the_exergy_of_cold_Theory_and_practical_examples
Physical exergy of a mass flow with p < p0 and/or T < T0 is the second part we need to check.

>     * sink: If a sink represents an exergy loss, it is regarded as E_D of the sink at the moment. That is not right. An exergy loss is only associated with the network. So, I would suggest that E_D = np.nan always and the exergy loss value is added to E_L of the network.

I named it :code:`E_D` for consistency with the other components, it is dealt in the desired way here: https://github.com/oemof/tespy/pull/215/files#diff-34b47c659ed2f64a0a185a3a644aed89ed89e20c01db2e966c4d1e5d2af126cdR2634-R2636. But we could name the attribute accordingly here.

>     * merge: According to "Thermal design and optimization" by Bejan et al on page 152 it is also correct to say: E_P = E_out and E_F = sum(E_in)

good!

>     * exergy balance of network: Right now E_D_network is the sum of E_D_comps. But we have to calculate the Network exergy balance separately. So E_D_network = E_F_network - E_P_network - E_L_network. Then we can print E_D_sum of components again as a value to compare with E_D_network.

This is already done here: https://github.com/oemof/tespy/pull/215/files#diff-34b47c659ed2f64a0a185a3a644aed89ed89e20c01db2e966c4d1e5d2af126cdR2690-R2698

src/tespy/networks/networks.py
```diff
@@ -2523,6 +2525,178 @@ def postprocessing(self):
...
+                    else:
+                        cp_E_P = cp.E_P * cp.calc_bus_efficiency(b)
+                    cp_E_D = cp_E_F - cp_E_P
+                    cp_epsilon = cp_E_P / cp_E_F
+
+                    if b in E_F:
+                        if b.comps.loc[cp, 'base'] == 'bus':
+                            self.E_F += cp_E_F
+                        else:
+                            self.E_F -= cp_E_P
```
juliusmeier, line 2666: Why is it cp_E_P and not cp_E_F ?

fwitte: Fuel exergy is inputs minus respective outputs (e.g. in a clausius rankine heat pump the compressor electrical power minus turbine electrical power). Therefore we need E_F for all power inputs and E_P for all power outputs and analogously for the E_P busses.

fwitte:
There is one more thing I want to highlight regarding the allocation of exergy of a mass flow and one more question regarding exergy loss.

- E.g. talking about the expansion of air from a compressed air storage: By definition the mass flow leaving the plant into the ambient after the turbine is an exergy loss of the network. But, one could also consider the mass flow leaving the plant as counterpart to a mass flow going into the plant from the storage representing fuel exergy. I checked the results of both mehotds, the exergy balance remains closed. But, the exergy efficiency of the network depends on the allocation of the exergy loss (reduction of fuel exergy or exergy loss). Therefore only the first method (allocate outflows to the exergy losses by specifying `exergy='loss'` for sinks) should be applied and we should highlight this in the documentation

- According to https://doi.org/10.1016/j.energy.2006.07.002 ("[...] exergy losses (E L ) associated with the transfer of exergy (through material and energy streams) to the surroundings.") exergy losses can also appear in dissipative heat exchangers, do I understand that correctly?

fwitte:
Going to merge this, remaining features will be implemented in a different PR.

PR #222, opened 2020-10-17
Pygmo optimization example added pygmo installation notes 

MarBrandt:
Adding a short description of installation methods for PyGMO using conda and pip.


fwitte:
Thank you for the suggestion, merged.

PR #226, opened 2020-10-19
Allow using Cuda for matrix inversion

jbueck:
At least when running rather large networks matrix inversion draws a lot of performance. This allows the use of Cuda for that operation to let it be run on the GPU, which improves performance quite a lot. As this is only possible for Nvidia GPUs and will not be useful for everyone it is only applied when the variable "use_cuda" in "global_vars.py" is set to true. So is the import of the package cupy, that is used for addressing Cuda.


fwitte:
Thank you for your contribution. This looks very interesting/promising. I would like to add the following information before we merge the new feature:

- benchmark tests, to identify in which cases usage of cuda is (most) beneficial.
- documentation update, so users will know about the feature. The documentation should be updated in the tespy_modules chapter within the section regarding the network solving. It may also be interesting to have a separate section dedicated to calculation speed improvements.

Unfortunately I cannot use the cuda feature as I do not work on a machine with a NVIDIA GPU. Therefore I would like to ask you, if you can perform the calculations on your machine. Can we have a call on this topic? I will try to come up with a testing/evaluation code for different use cases.
Best

fwitte:
And we need installation instructions for cuda/cupy, as `pip install cupy` does not work (at least for me). You have to install cuda first, I guess?

jbueck:
Yes, sure, I can do that. The performance improvement is best when running tespy in parallel processes, as the CPU load seems to be better balanced between the processes and cores, while the GPU is working. We could try to show that as well.

jbueck:
> And we need installation instructions for cuda/cupy, as `pip install cupy` does not work (at least for me). You have to install cuda first, I guess?

For installing cupy using pip, some prerequisites have to be met, which I think depend on the OS you are using. That is explained on the docs page: [https://docs.cupy.dev/en/stable/install.html#installing-cupy](https://docs.cupy.dev/en/stable/install.html#installing-cupy)  There are also precompiled binary packages that can be installed using pip, which makes it easier to get it running. Of course you have to have Cuda installed and working for the package to run. Because of that I put the import cupy statement in that if-condition, so that you only have to install cupy if you really want to use it.

fwitte:
Benchmark examples:

- [ ] simple network with max 15 components and one pure fluid only
- [ ] simple network with max 15 components using gas mixtures
- [ ] simple network with max 15 components using pure fluids and gas mixtures simultaneously
- [ ] large network with max 30 components, single fluid only
- [ ] large network with max 30 components, pure fluids and gas mixtures simultaneously
- [ ] huge network with > 50 components, single fluid only

jbueck:
These are the results of the test cases evaluating the activation of CUDA and always computing all equations.

[results_overview.zip](https://github.com/oemof/tespy/files/5618153/results_overview.zip)

PR #227, opened 2020-10-20
Adjust some component value limits

jbueck:
Adjust limits for zeta max value and for ks if using HW equation. Value limits are changed for simple heat exchanger and valve.


fwitte:
Hi @jbueck,
thanks for your contributions and suggestions. We can quickly merge this one, the other PRs require some updates in the documentation. I'll comment there respectively.
Best
Francesco

PR #228, opened 2020-10-20
Remove conditions for computing equations

jbueck:
Remove the conditions that trigger whether an component equation is to be computed. In my simulations this often caused fluid states to be pushed out of bounds and attribute values to be out of their limited range when being calculated in calc_parameters after a successfully run simulation. Removing the calculation conditions solved those problems for me.

### DIFF
--- a/docs/whats_new/v0-3-5.rst
+++ b/docs/whats_new/v0-3-5.rst
@@ -35,6 +35,17 @@ Other Changes
 
       mynetwork.print_results(colored=False)
 
+- Add a flag to activate calculation of all component equations in every
+  iteration. This improves stability in some cases but reduces calculation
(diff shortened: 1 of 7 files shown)


fwitte:
Is this the case for initial calculations or does it happen in any kind of calculation? I suggest to make the feature optional, so we can preserve the speedup for networks, where this is not an issue. Similar to the cuda/cupy implementation

jbueck:
> Is this the case for initial calculations or does it happen in any kind of calculation? I suggest to make the feature optional, so we can preserve the speedup for networks, where this is not an issue. Similar to the cuda/cupy implementation

I ran into this quite som times, but yes, let's do it optional so people can decide according to their needs. I will push a commit for that.

fwitte:
Moved to #226

PR #233, opened 2020-11-04
checking fluid vector for fluids not within the network's fluid list

### LINKED ISSUE #232: No error is raised, if a fluid vector contains a fluid not within the network's fluid list
See the code in #231

MarBrandt:
- checking fluid vector for fluids not within the network's fluid list
- added 'def test_bad_fluids_in_fluid_vector' in 'test_errors.py'
- resolve #232

### DIFF
--- a/docs/whats_new/v0-3-5.rst
+++ b/docs/whats_new/v0-3-5.rst
@@ -35,7 +35,12 @@ Other Changes
 
       mynetwork.print_results(colored=False)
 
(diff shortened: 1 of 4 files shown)


fwitte:
Hi, thanks for the PR! We need to make two adjustments:
- one test is failing with the wrong error triggered due to the new feature. This means, the error message, that originally should have been triggered might be obsolete or we need to adjust the test.
- add short description in what's new.

Best regards

PR #236, opened 2020-11-16
Fix: General documentation updates and fixes

fwitte:


fwitte:
Changes from 854facb have been reverted!

PR #238, opened 2020-12-08
Add new Feature: IF97 implementation (with a fix)

fwitte:
This PR introduces IF97 of water with a small fix.


fwitte:
The IF97 implementation is fixed as described in discussion https://github.com/CoolProp/CoolProp/issues/1918.

PR #239, opened 2020-12-14
Fix/improve the exergy balancing methods of components

fwitte:
This PR adjusts the exergy balancing methods introducing allocation of thermal and mechanical exergy splitting in physical exergy.


fwitte:
- Merging this into dev as some features are required for different feature.
- What's New will be updated later.
- Implementations may be subject to changes, as some issues still persist.

PR #242, opened 2021-01-04
Prompt error message, if network is missing required fluids for combu…

### LINKED ISSUE #241: When using the characteristic line, a warning always appears, and it cannot operate normally!
Hey Francesco,
When I created my own thermal system, the gas turbine parameters were set by referring to the chp example, which is like this:`com1.set_attr(pr=14,eta_s=0.91,design=['pr','eta_s'],offdesign=['eta_s_char'])
`
However, when I run, the following warning always appears:
@LREH4M%2KPYBV$(_YUTV~5
Currently I do not have my own data, and I need to use the default feature line. So I want to know whether the appearance of this warning will affect the calculation and the appropriate solution?
Thank you! Looking forward to your reply!

fwitte:
Linked to #241

### DIFF
--- a/docs/whats_new/v0-3-5.rst
+++ b/docs/whats_new/v0-3-5.rst
@@ -56,8 +56,7 @@ API Changes
 New Features
 ############
 - Add methods for exergy and entropy analyses of networks.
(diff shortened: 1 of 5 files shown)


fwitte:
Resolved by https://github.com/oemof/tespy/commit/7adff43

PR #245, opened 2021-02-05
New Feature: User defined equations

fwitte:
This PR introduces generic user defined equations that can be applied within your model.


fwitte:
- [ ] Change from UserDefinedFunction to UserDefinedEquation
- [ ] Add some text in the online documentation

PR #251, opened 2021-03-01
Add new Feature: Sankey diagrams for exergy flow plotting

fwitte:
This PR adds a data generator for the [plotly](https://plotly.com/python/sankey-diagram/) library to use sankey diagrams.

An automatically generated diagram could look like this:

newplot(5)

To finish up:

- [ ] Add test for prohibited fkt_group labels
- [ ] Add test with large power plant
- [ ] Add tests for each of the exergy balances in the components
- [ ] Add a documentation section for analyses module
- [ ] Update What's New


fwitte:
To finish up:

- Add test for prohibited fkt_group labels
- Add test with large power plant
- Add a documentation section in the TESPy modules overview
- Update What's New

src/tespy/tools/analyses.py
```diff
@@ -0,0 +1,759 @@
...
+
+        - The results are sorted beginning with the component having the
+          biggest exergy destruction by default.
+        - Components with an exergy destruction smaller than 1000 W is not
+          printed to prompt by default.
+
+        Parameters
+        ----------
+        E_D_min : float
+            Minimum exergy destruction to be printed to prompt.
```
fwitte, line 698: @juliusmeier, we might want to remove this entry as the user can simply save the DataFrame into his variable space for further processing and there is no necessity to have this in the code. What do you think?

src/tespy/tools/analyses.py
```diff
@@ -0,0 +1,759 @@
...
+            colors['E_F'] = rgba_list[1]
+            colors['E_D'] = rgba_list[2]
+            colors['E_L'] = rgba_list[3]
+            i = 4
+            for f in self.nw.fluids:
+                colors[f] = rgba_list[i]
+                i += 1
+            for b in self.internal_busses:
+                colors[b.label] = rgba_list[i]
+                i += 1
```
fwitte, line 662: Generate the dictionary, and update with the passed dictionary afterwards. By this, the user does not need to specify a full dictionary, but only the colors he wants to change.

src/tespy/tools/analyses.py
```diff
@@ -0,0 +1,759 @@
...
+                data['value'].abs() >= display_thresold]
+
+        self.remove_transit_groups(group_data)
+
+        if len(node_order) == 0:
+            node_order = (
+                ['E_F'] + [b.label for b in self.E_F] +
+                [fkt_group for fkt_group in self.group_overview.index] +
+                [b.label for b in self.internal_busses + self.E_P + self.E_L] +
+                ['E_P', 'E_L', 'E_D'])
```
fwitte, line 636: Node order should be checked for exact length. If length is larger than zero but smaller than total amount of nodes, an error should be thrown.

src/tespy/tools/analyses.py
```diff
@@ -0,0 +1,759 @@
...
+            index=['E_F', 'E_P', 'E_D', 'E_L'], dtype='float64')
+        self.network_data[:] = 0
+
+        # physical exergy of connections
+        for conn in self.nw.conns['object']:
+            conn.get_physical_exergy(pamb_SI, Tamb_SI)
+            self.connection_data.loc[conn.label] = [
+                conn.ex_physical, conn.Ex_physical]
+
+        self.group_data = {}
```
fwitte, line 350: Rename group_data to sankey_data and group_overview to group_data

src/tespy/tools/analyses.py
juliusmeier: Good point, I agree.

docs/analyses/exergy.rst
```diff
@@ -6,15 +6,32 @@ Introduction
...
+But regarding the design and analysis of thermal systems, the idea that something can be destroyed is useful. 
+According to the second law of thermodynamics, the conversion of heat and internal energy into work is limited. 
+This constraint and the idea of destruction are applied to introduce a new concept: exergy. 
+Exergy can be destroyed due to irreversibilities and is able to describe the quality of different energy forms. 
+The difference in quality of different forms of energy shall be illustrated by the following example. 
+1 kJ of electrical energy is clearly more valuable than 1 kJ of energy in a glass of water at ambient temperature. :cite:`Bejan1996`
+In literature, exergy is defined as follows:
+“The exergy of a system is the maximum theoretical useful work (wave work or electrical energy) 
+that is obtained when the system is brought into complete thermal, mechanical and chemical equilibrium 
+with the thermodynamic environment and there are interactions only with this environment. “ :cite:`ET1`
```
fwitte, line 32: Can we find a different source, that is publicly available?

juliusmeier: Ah right, I will do some research

PR #273, opened 2021-06-17
Add new feature: nDodecane as fuel for combustion

### LINKED ISSUE #272: Add n-Dodecane to combustion chamber component
Hi!

I'm trying to model an aviation jet-prop, which uses Jet A-1 fuel, using TESPy.

Using n-Dodecane C12H26 instead of Jet A should be good enough, and it's included in CoolProp, however the combustion chamber component does not support it.

I've tried to add it myself, by adding it to the list of allowable fuels in combustion_chamber.setup_reaction_parameters, and adding the enthalpy of formation `hf['Dodecane'] = -352.1` to combustion_chamber.calc_lhv. The LHV calculation was pretty good, got ~44 MJ/kg.

When running a simple model with source --> compressor --> combustion chamber --> sink, the calculation can't seem to converge, no matter the starting conditions, so I guess I'm missing something else (the same model works with CH4 just fine).

What else is missing that will enable me to use this fuel in the model?

fwitte:
Resolve #272 

This PR adds new fuels to the class CombustionChamber. To check:

- Is the energy balance calculated correctly even for liquid fuels?
- Default starting values for nDodecane are really bad.

### DIFF
--- a/docs/whats_new/v0-4-4.rst
+++ b/docs/whats_new/v0-4-4.rst
@@ -1,0 +1,25 @@
+v0.4.4 - Someversion (July, 12, 2021)
++++++++++++++++++++++++++++++++++++++
+
+Documentation
+#############
+- Fix some typos here and there.
+
+Bug Fixes
+#########
+- Fix pandas version 1.3.0 dependencies
+  (`PR #277 <https://github.com/oemof/tespy/pull/277>`_).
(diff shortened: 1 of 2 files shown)


fwitte:
Will add nDodecane with gaseous reference state for the moment. The topic of the fluid's reference state for combustion calculations may need in depth investigation in the future.

PR #294, opened 2021-11-30
New Feature: Support for INCOMP binary mixtures

jelic1marko:
This pull request aims to implement minor changes to the backend processing of fluid properties so that mass-based binary mixtures included in CoolProp can be natively supported as network fluids. These fluids are crucial for simulation of low temperature applications such as ground-sourced energy.

The API is relatively straightforward. These mixtures should be included into the network just as any other, with a specified mass fraction and back end following the given syntax:

nw = Network(fluids=['INCOMP::MEG[0.2]', 'INCOMP::MPG[0.4]', 'water', ...], ...)

while these fluids can later be utilized as connection parameters like:

meg_in_hx.set_attr(fluid={'MEG[0.2]': 1, 'MPG[0.4]': 0, 'water': 0, ...}, ...).

No bugs appear to arise from this introduction and the regular operation appears not to be influenced in any way.

A simple demonstration of how these mixtures can be utilized is depicted in tutorials/binary_mixtures.py.

### DIFF
--- a/docs/tespy_modules/fluid_properties.rst
+++ b/docs/tespy_modules/fluid_properties.rst
@@ -45,8 +45,10 @@ cover liquid state only.
 
 Fluid mixtures
 --------------
-CoolProp provides fluid properties for two component mixtures. BUT: These are
-NOT integrated in TESPy! Nevertheless, you can use fluid mixtures for gases.
+CoolProp provides fluid properties for two component mixtures. Through the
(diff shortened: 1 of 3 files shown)


fwitte:
Thank you very much for you suggestion :). Will look into it as soon as possible. We might need to set up one or two tests, and then we can include it with the next release.

fwitte:
Dealt with by #384

Issue #299, opened 2022-01-10
New component: Diabatic Compressor


maltefritz:
For the simulation of heat pumps, we (@jfreissmann and me) would like to use an actively cooled compressor. With the current implementation, this is not possible, because an adiabatic compression is assumed within the standard component.

Together with @fwitte we implemented a version of the component that uses a fixed ratio between the cooling heat flow and compressor power. This ratio currently has to be set by the user. Our aim is to have this ratio be variable or even a result of other component parameters (i.e. polytropic efficiency). You can find our code in the feature branch features/DiabaticCompressor.

Questions to other users working on similar projects:
    - Does anybody have literature regarding the stationary modelling of diabatic compressors?
    - Has anybody already implemented a diabatic compressor or worked on a similar problem?

PR #300, opened 2022-01-10
New Component: DiabaticCompressor

maltefritz:
See issue #299


fwitte:
Is this still ongoing?

fwitte:
One could implement polytropic compressor here, for example see [doi.org/10.1115/1.3239827](https://doi.org/10.1115/1.3239827) or https://www.turbomachinerymag.com/view/compressor-performance-and-thermodynamic.

PR #326, opened 2022-03-24
Include chemical exergy in ExergyAnalysis

KarimHShawky:
Addition to the chemical exergy feature to be able to use the calculated chemical exergy in the ExergyAnalysis

- Added Ex_chemical to E_bus in Components Sink and Source.
- Added Exergy Balance to Diabatic Combustion Chamber.
- Modified ExergyAnalysis to include chemical Exergy. Work in Progress

Currently the fkt_groups in the ExergyAnalysis are still not working properly. There doesn't exist a way to Split into chemical and physical exergy when assigning the E_P and E_F of the Network 
=> epsilon therefore can differ


fwitte:
Closing this, we move on in #322

PR #327, opened 2022-04-08
Feature/thermosteam

nkawerau:
This feature provides the option to use Thermosteam (https://thermosteam.readthedocs.io/en/latest/) as fluid properties back end, located in tespy/tools/fluid_properties.py. Use the keyword "TS" for Thermosteam instead of the standard usage of "HEOS" (Helmholtz equation of state).

Example: `fluid_network = Network( fluids = ["TS::H2O", "TS::KOH"] )`

For now, the feature is designed to use aqueous potassium hydroxide [H20(l) + KOH(aq)] as liquid fluid components and hydrogen + oxygen [H2(g) + O2(g)] as gaseous fluid components, a more generalized approach is targeted.


fwitte:
This will fall into works of #384, closing here for now.

PR #335, opened 2022-04-29
fix electrolyzer documentation

BenPortner:


src/tespy/components/reactors/water_electrolyzer.py
```diff
@@ -130,8 +131,8 @@ class WaterElectrolyzer(Component):
     ... WaterElectrolyzer)
     >>> from tespy.connections import Connection
     >>> from tespy.networks import Network
-    >>> from tespy.tools import ComponentCharacteristics as dc_cc
     >>> import shutil
+    >>> import numpy as np
```
fwitte, line 135: Thank you for noticing this, I wonder how the tests passed with the missing import... In this example it is only used to unset the volumetric flow of hydrogen to the compressor (L181/L182), I think I will remove it and use None instead. Originally, `np.nan` was used to unset values, `None` was introduced later. It seemed more convenient to me, since you do not necessarily need to import 

> numpy.

PR #336, opened 2022-04-29
add electrolyzer example

BenPortner:
took the example from the doc-string and added it as a stand-alone example to the "example/tutorials" section of the documentation

### DIFF
--- a/docs/tutorials_examples/electrolyzer.rst
+++ b/docs/tutorials_examples/electrolyzer.rst
@@ -1,0 +1,72 @@
+Water Electrolyzer
+---------------
+
+This example shows how you can use the water electrolyzer component in TESPy.
+
+.. figure:: _images/WaterElectrolyzer.svg
+    :align: center
+
+    Figure: Topology of the water electrolyzer.
+
+The water electrolyzer produces hydrogen and oxygen from water and power.
+
+Example
+#######
+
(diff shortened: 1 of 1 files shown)


fwitte:
Hi,

thank you for your suggestion and please excuse for coming back so late to it. Generally, the docstring examples are accessible in the API documentation. However, they might be kind of hidden (https://tespy.readthedocs.io/en/main/api/tespy.components.html#tespy.components.reactors.water_electrolyzer.WaterElectrolyzer). I think it would be better, if there is a single place for each example.

Maybe we can add a more exhaustive example in the examples list. Maybe something with a storage integration or similar. I will put it up on the oemof meeting agenda for next week. Maybe a small break out group is willing to develop an example :). It would be helpful in any way, I think! If you have an example implementation ready yourself, you are welcome to share it.

Thanks again, I will come back to the PR after [next week's user meeting](https://oemof.org/2022/04/05/next-oemof-user-meeting-in-may-2022-in-flensburg/#more-3337).

fwitte:
Hi @BenPortner,

I'll close this for the moment, since I restructured many things in the software. Still like the idea to have an example for the hydrogen feature on the top level of the docs, but I would like it to be  a slightly larger one... Thank you still!

Best

Francesco

PR #346, opened 2022-06-09
Starting Values Tutorial

maltefritz:
Add tutorial for the generation of starting values through the example of a single stage heatpump with internal heat exchanger.


fwitte:
Hi,

thank you for your suggestion, very good addition!

Could you please also update the figure (following the template I sent you) and add lables within the figure following this structure: https://github.com/fwitte/refrigeration_cycle_exergy/

The main cycle connections are 0, 1, 2, 3, ..., (0 if a cycle closer is included) the next system starts with 10/11 (10 if a cycle closer is included), 11 if not. The same needs to be applied in the lables of the Connections in the python script. Thanks a lot :)

fwitte:
I updated the description in the docs, could you please make another spell check? Some of the code needs minor modifications as well, I think it is best, if you have a meeting on this.

fwitte:
@maltefritz: I created a jupyter notebook because I want to replace the .rst files in the tutorials/examples section in the future. Could you make a spell check inside of that? I will try to figure out, how to integrate the figures and then, how to make readthedocs display the notebook correctly. Thank you :)

PR #384, opened 2023-01-29
Make the fluid property back-end flexible

fwitte:
There is a lot of issues with the fluid property back-end that hinder (easy) integration of new features, such as the incompressible mixtures, all sorts of CoolProp inbuilt mixtures and stuff like absorption heat pumps etc.. This PR will make a lot of breaking changes to the current structure but provide an easily adaptable back-end later.


fwitte:
A first small test case is available in the `new_fluid_properties_tests.py` file:

- It features a network with two single (non related) connections. The example implementation uses the new fluid_properties module structure.
- The changes allow to reduce the Jacobian matrix by removing trivial variables from the connections (specified number for mass flow is not a variable).
- This also removes the need to calculate derivatives towards variables that cannot change their value.
- The connections class will hold the connections equations in a similar fashion as the components do.

fwitte:
Tests now run all, this is completed.

PR #390, opened 2023-02-13
Feature/new logger

jowr:
**General**

The proposed changes introduce a constant string id for the logger object that is retrieved from the logging module. This allows us to modify the logging functionality in a programmatic way and simplifies the work that has to be done in libraries and applications that consume the TESPy library. 

New logging levels have been introduced to facilitate a structured progress reporting and result handling in the future. This is going to be covered in a separate PR since some more changes in the network solver would make it hard to review both changes simultaneously.

Shorthand functions have been added to the `tespy.tools.logger` file that mimic the behaviour of the original logging functions. All source code files using the logging facilities have been altered to use the new functions. The shorthand functions modify the stack handling in the logging library to make it trace past the wrapper functions in the `tespy.tools.logger` file. This particular remains to be tested, but it seems to work at a first glance.

**Documentation**

All new functions have local documentation attached. Please let me know, if more documentation is needed other places.

**Testing**

The existing tests use the logging functionality and it should thus be regarded as tested.

### DIFF
--- a/src/tespy/tools/logger.py
+++ b/src/tespy/tools/logger.py
@@ -20,6 +20,9 @@ TESPY_LOGGER_ID = "TESPyLogger"
 TESPY_PROGRESS_LOG_LEVEL = logging.DEBUG + 1 # 11
 TESPY_RESULT_LOG_LEVEL = logging.DEBUG + 2 # 12
 
+# Capture warnings globally instead of per file
+logging.captureWarnings(True)
+
 # Create a bunch of shorthand functions, this is mostly 
 # copied straight from the logging module.
 def getLogger():
--- a/src/tespy/tools/logger.py
+++ b/src/tespy/tools/logger.py
@@ -17,7 +17,106 @@ from logging import handlers
 import tespy
 
 TESPY_LOGGER_ID = "TESPyLogger"
-TESPY_PROGRESS_LOG_LEVEL = 31
+TESPY_PROGRESS_LOG_LEVEL = logging.DEBUG + 1 # 11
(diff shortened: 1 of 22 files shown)


jowr:
closes #390

src/tespy/tools/logger.py
```diff
@@ -16,6 +16,259 @@
...
+
+    To pass exception information, use the keyword argument exc_info with
+    a true value, e.g.
+
+    debug("Houston, we have a %s", "thorny problem", exc_info=1)
+    """
+    # We force the logging framework to trace past this file
+    if "stacklevel" not in kwargs:
+        kwargs["stacklevel"] = 1
+    kwargs["stacklevel"] += 1
```
fwitte, line 62: This specific piece could be a function, since it is used in all logging functions. Or we can use `kwargs.get("stacklevel", 1) + 1`.

fwitte:
@jowr, thank you very much for the suggested modifications and additions. I really do like them. I have modified some small things locally and wanted to push them to your branch but I have no permissions to do so (On merge requests you can allow maintainers to push to the branch. I really guess you know that, so maybe it was intentional? ).

I am just going to push those commits to my dev and then merge your branch.

src/tespy/tools/logger.py
fwitte: I discovered that was incorrect, I implemented it however in an own function.

jowr: Thank you for reviewing that part.

jowr:
Thanks again and I am going to keep an eye on the flake8 formatting myself next time 😉

PR #391, opened 2023-02-14
Report progress and results with the logger object

jowr:
**General**

This PR is a follow-up on #390 - it should be merged after #390 has been integrated.

The main purpose of this PR is to silence stdout when running TESPy. For compatibility reasons, the default settings make the system run as before, but it is now possible to send iteration progress reports via the logging framework. The same is the case for the results. 

**Documentation**

Some internal functions have been renamed, but the external interface of TESPy is unchanged. The only noticeable change is that the iteration output now contains a progress indicator.

### DIFF
--- a/src/tespy/networks/network.py
+++ b/src/tespy/networks/network.py
@@ -1545,7 +1545,7 @@ class Network:
 
     def solve(self, mode, init_path=None, design_path=None,
               max_iter=50, min_iter=4, init_only=False, init_previous=True,
-              use_cuda=False, always_all_equations=True):
+              use_cuda=False, always_all_equations=True, print_result=True):
         r"""
         Solve the network.
 
@@ -1590,6 +1590,10 @@ class Network:
             Calculate all equations in every iteration. Disabling this flag,
             will increase calculation speed, especially for mixtures, default:
             :code:`True`.
+        
+        print_result : boolean
+            Print the results to the terminal instead of returning a string 
+            with the results, default: :code:`True`.
 
         Note
         ----
@@ -1659,7 +1663,7 @@ class Network:
         logger.info(msg)
 
         self.solve_determination()
-        iter_result = self.solve_loop()
+        iter_result = self.solve_loop(print_result)
 
         if self.lin_dep:
             msg = (
@@ -1693,7 +1697,7 @@ class Network:
         logger.info(msg)
         return iter_result
 
-    def solve_loop(self):
+    def solve_loop(self, print_result=True):
         r"""Loop of the newton algorithm."""
 
         result = ""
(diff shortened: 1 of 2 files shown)


fwitte:
Hi @jowr, @Vily-ipu,

thanks for your efforts and contribution, really appreciate that!

I already merged #390, but something is weird here. Your commits up until 37a5abfb075e76373088ce1150a93090e582dd6c are all in dev, but git does not really recognize that... I need to figure out what is wrong once the PR is ready. Maybe it will be resolved by merging the current dev into this branch. Just let me know, once you want to move forward!

Thanks and have a sunny week!

src/tespy/networks/network.py
```diff
@@ -2698,7 +2698,7 @@ def print_results(self, colored=True, colors={}, return_result=False):
                         )
 
         # connection properties
-        df = self.results['Connection'].loc[:, ['m', 'p', 'h', 'T']]
+        df = self.results['Connection']
```
fwitte, line 2687: There will be some small adjustment needed: Without unselecting the fluid names, there will be an error later in the displaying the results in line 2709.

jowr: Thank you - we can just roll this commit back.

fwitte:
@jowr: you can run python -m tox -e check to make the flake testing and everything locally, if you installed the dev dependencies of TESPy

jowr:
Thanks, unfortunately I cannot run the tests locally because our project folders have spaces in their paths and that breaks tox...

jowr:
Hi @fwitte - Is there anything more I can do? I think the failing test is not related to the changes in this PR.

fwitte:
@jowr, thank you for reminding me and sorry for my delays. I am in holidays right now, I will scoop up the open points next week.

jowr:
No worries - I was just wondering whether I overlooked a change request.

fwitte:
Got it, thank you.

For the moment I'll take care about those tests and I will make some minor adjustments in style. There should be nothing to do from your side for now.

fwitte:
@jowr: Merged, thanks again!

fwitte:
@jowr: On doing some more things with this merged into dev, I just noticed, that the progressvalue sometimes does not seem to make too much sense. Or I have some misunderstanding... I will try to get some time next weekend to look into it, and I may come back to you for some questions later :).

jowr:
That is correct - I currently only look at the stop criterion and assume that the solver progresses exponentially (linearly in terms of the log of the residual value). This is more an indicator than a real value. If you think this confuses more than it helps, we can remove it again. However, I think it would be super cool if we could come up with a correct definition of the progress value...

PR #443, opened 2023-11-05
Add sorption features to TESPy

fwitte:
This PR will add equations to the fluid property back-end with the aim to model absorption and desorption processes, e.g. in sorption heat pumps or refrigeration applications. The initial drafts can be found in the root directory of the repository (test_absorber.py and test_desorber.py). The context is based on the examples at https://github.com/LudwigIrrgang/SorpPump.


fwitte:
An example can be found here currently: https://github.com/fwitte/tespy-sorption-models

fwitte:
Will be closed in favor of a different approach

PR #474, opened 2024-01-12
add exergoeconomic analysis

FriederikeTUB:


fwitte:
Hi @FriederikeTUB,

there is a new software tool in the oemof world, exerpy: https://github.com/oemof/exerpy. In incorporates the concepts built by you, and I will deprecate the exergy analysis in tespy soon (see #691). Thank you again for the contributions, they will live in a different place and this will not be merged here.

Best

Francesco

PR #502, opened 2024-04-15
Merge commits from ipudk/tespy

fwitte:
I tried working with git-filter-repo to separate the core code from the application specific new parts to split the changes into two repositories, which was unsuccessful. There seems to be something really strange in the history of the ipudk fork, which is why I ended up cherry picking the commits and adding a commit on top of that deleting the respective files, which would not be part of the core.


fwitte:
There is many things to process here, so it might take quite a bit of time to complete the PR.

PR #504, opened 2024-04-16
Addition of some features implemented by ipudk

fwitte:
PR in favor of #502


fwitte:
Some (not all of them) of the features have been ported to the main branch already, this will not be progressing but might be a starting point for reimplementation of missing features from current main.

PR #507, opened 2024-04-29
Improve writing JSON file method.

jfreissmann:
**General**

When writing JSON data to a file, the `json.dumps` method is used to create a string in JSON format, which is then written to disk with the `write` method of the `file` object. The extra conversion is not necessary, since the `json.dump` method directly writes the data to the file, without the extra step of converting the data to a string. See for example [this comparison on StackOverflow](https://stackoverflow.com/questions/36059194/what-is-the-difference-between-json-dump-and-json-dumps-in-python) or the [json module documentation](https://docs.python.org/3/library/json.html).


fwitte:
Thanks a lot for the contribution! :)

PR #515, opened 2024-05-26
Implement a moving boundary heat exchanger for improved part load modeling

fwitte:
Especially in heat pump or refrigeration machine applications, the condenser component may have a significant heat transfer share from the desuperheating of the working fluid. The heat transfer coefficients for desuperheating and condensation usually differ strongly, therefore using an overall `kA` (`UA`) value for the complete process to model part load may not be correct. Therefore an approach is implemented with a moving boundary, which will split the virtual area of the condenser into a desuperheating and a condensation part. The split will happen proportionally to the heat transfer of each subprocess and be validated using reference data.


fwitte:
Another interesting approach would be following this publication: https://www.sciencedirect.com/science/article/pii/S1359431114011570

fwitte:
This will serve as basis for new heat exchanger models. Another one, i.e. a 1-D heat exchanger including pressure losses will follow.

PR #521, opened 2024-06-12
Extend get_plotting_data method of Drum class.

jfreissmann:
**General**

The `get_plotting_data` method of the `Drum` class only returns one process line connecting the saturated liquid `'out1'` with the saturated vapor `'out2'` so far. This can work sometimes, because the input saturated liquid-vapor mixture falls somewhere between these points, but the desuperheating of the gas `'in2'` is lost and can not be shown. Unfortunately, the `Drum` component does not lend itself to a clear fix that is in line with other components and the general discription of the method. That is, because changing the returned dict to contain one line from `'in1'` to `'out1'` and one from `'in2'` to `'out2'` loses the depiction of the isolation of the saturated vapor from the mixture. Therefore, I propose this solution connecting every inlet to every output and letting the user decide which lines they want to plot.

What do you think about this? Either way, in my opinion the old version can not stay, as you can not show the desuperheating. Additionally, the method desciption did not match the return value in the case of the `Drum`.

### DIFF
--- a/docs/whats_new/v0-7-5.rst
+++ b/docs/whats_new/v0-7-5.rst
@@ -7,7 +7,13 @@ Documentation
   process into the results.
   (`PR #514 <https://github.com/oemof/tespy/pull/514>`__).
 
+Other Features
+##############
+- More isolines are now available for the drum
+  (`PR #521 <https://github.com/oemof/tespy/pull/521>`__).
+
 Contributors
 ############
 - Francesco Witte (`@fwitte <https://github.com/fwitte>`__)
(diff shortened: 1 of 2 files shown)


fwitte:
Hi @jfreissmann,

thanks a lot for your contribution. I will look into the changes later and let you know how we proceed:).

Best

fwitte (changes requested):
Looks good, thank you! Can you change it, so the key `1` remains the old version (outlet 1 to outlet 2). We can add a warning message, that this behavior will change in the future and change it with the next major release.

Issue #586, opened 2024-12-28
Bug in post-solve calculation of ttd_min


tlmerbecks:
A typo has crept into the calculation of ttd_min in the calc_parameters function of the base heat exchanger class

fwitte:
Resolved in #587, thank you!

PR #587, opened 2024-12-28
Fixed bug

tlmerbecks:
In the post-solve calculation of ttd_min (see calc_parameters() ) there was a typo, resulting in the ttd_min being calculated as the minimum value of ttd_u and tdd_min (incorrect) instead of the minimum of ttd_u and ttd_l.

This fixes issue #586

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -3,6 +3,7 @@ What's New
 
 Discover notable new features and improvements in each release
 
+.. include::  whats_new/v0-7-8-002.rst
 .. include::  whats_new/v0-7-8-001.rst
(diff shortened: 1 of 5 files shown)


fwitte:
@tlmerbecks, thank you for reporting and fixing the bug.

On top of that, I conceptualized and implemented an idea for the component testing, which should make it easier to catch these kinds of issues in the future. I did not fully push it through all components, but I will still merge this and make a bugfix release. The idea also needs some refining in my opinion... While doing so, I found, that the same bug was present in the `Condenser` class and modified that as well.

Best

Francesco

PR #592, opened 2025-01-05
Vapour Quality and State Reporting

tlmerbecks:
Ciao @fwitte,

I noticed that the property reporting for `Connections `does not include the vapour quality or the state, hence the overall aim of this pull request is to include them in the table of properties for `Connections`.

I implemented the following changes:

- Aligned the calculation of the vapour quality for the `CoolPropWrapper` with that of the `IAPWSWrapper`. Currently the `CoolPropWrapper` reports a vapour quality of `-1` for all single phase states (irrespective of it being gas or liquid).  With the changes below, the vapour quality is now reported as `0` for liquids and `1` for gases, for subcritical conditions.
- Added a state_ph function to the base `FluidPropertiesWrapper`, the `CoolPropWrapper `and the `IAPWSWrapper`. Unfortunately, I lack expertise with Pyromat and so have not done this... yet)
- Extended the `fluid_properties_data` (global_vars) with a `state` properties field
- Added an exception rule for the unit conversion of the new `state` property
- Added the `vapour quality` and `state` to the reported `Connection` results

I hope the above is clear enough. Is there anything else that would be helpful?

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -3,6 +3,7 @@ What's New
 
 Discover notable new features and improvements in each release
 
+.. include::  whats_new/v0-7-9.rst
 .. include::  whats_new/v0-7-8-002.rst
 .. include::  whats_new/v0-7-8-001.rst
 .. include::  whats_new/v0-7-8.rst
--- a/docs/whats_new/v0-7-9.rst
+++ b/docs/whats_new/v0-7-9.rst
@@ -1,0 +1,14 @@
+v0.7.9 - Under development
+++++++++++++++++++++++++++
+
+New Features
+############
+- Implement a new property for connections to report the phase of the fluid,
+  i.e. :code:`"l"` for liquid, :code:`"tp"` for two-phase and :code:`"g"` for
+  gaseous. The phase is only reported in subcritical pressure
+  (`PR #592 <https://github.com/oemof/tespy/pull/592>`__).
+
+Contributors
+############
+- `@tlmerbecks <https://github.com/tlmerbecks>`__
+- Francesco Witte (`@fwitte <https://github.com/fwitte>`__)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ exclude = ["docs/_build"]
 
 [project]
 name = "tespy"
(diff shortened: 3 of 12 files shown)


tlmerbecks:
Looks like my PR was not overly successful... Unfortunately, I am somewhat struggling to understand exactly what tests have failed

fwitte:
@tlmerbecks, thank you for your contribution. I will have a look today evening, my first guess is, that it is because the `state` keyword is already utilized for different context (may be `'l'` or `'g'` for liquid/gas) in order to help convergence. If specified, the solver forces the state to be liquid or gaseous. Adding `'tp'` is a nice touch, but for the rest I have to look in detail into your changes. I'll leave you some feedback later :).

tlmerbecks:
Ciao @fwitte,

that may well be it... that being said, I had kind of hoped that with these additions the `state` variable could be evaluated mid- or post-solve. This would allow the state to be reported as part of the results table (similarly to the pressure, enthalpy, etc.) and/or be used in component calculations. (I am working on another pull request for a new `SteamTurbine` component that handles wet-expansion, as you would see it in geothermal or nuclear power plants, using an empirical correction for steam wetness).

tlmerbecks:
Actually, it seems as though the `state` variable is not declared via the dictionary generated by the `get_parameters` function, instead it is declared directly. However, with my additions, it would be simply overwritten when all other parameters are set.

```python

class Connection:
        def __init__(self, ...):

            # some other code

            # set default values for kwargs
            self.property_data = self.get_parameters()  # I defined the state field in here
            self.parameters = {
                k: v for k, v in self.get_parameters().items()
                if hasattr(v, "func") and v.func is not None
            }
            self.state = dc_simple()  # currently tespy declares the state variable here
            self.property_data0 = [x + '0' for x in self.property_data.keys()]
            self.__dict__.update(self.property_data)  # <-- overwrites the directly declared variable with my definition...

            # some other code

```

fwitte:
@tlmerbecks, thank you for your contribution! I updated the changelog, bumped the version and removed that commeted section in the `wrappers.py` `Q_ph` method of the `CoolPropWrapper`.

PR #602, opened 2025-01-24
Steam Turbine with wet expansion

tlmerbecks:
**General**

With this PR I would like to submit a new turbine component, which accounts for wet expansion effects using the empirical Baumann correlation.

The expansion of saturated steam (e.g. geothermal or nuclear power applications) inevitably enters the vapour dome and two-phase effects lead to a reduction in the isentropic efficiency. The empirical Baumann rule effectively corrects the dry isentropic turbine efficiency with the average wetness across the turbine.

* Implementation of the `SteamTurbine` component. I chose to make this a separate component (inheriting from `Turbine`) to avoid confusion - though the `SteamTurbine` behaves exactly like a `Turbine` unless the `eta_dry_s` attribute is set
* Updated the `CoolPropWrapper` `Q_ph` function so that it now returns a meaningful value for `Q` when the fluid is single phase (i.e. 0.0 for liquid and 1.0 for vapour)

**Documentation**

This is still work in progress... I was not quite sure what would be the mvp for this

**Testing**

This is still work in progress... I was thinking to shadow the testing for the regular `Turbine` component

Comments and suggestions are welcome :)

### DIFF
--- a/src/tespy/components/turbomachinery/steam_turbine.py
+++ b/src/tespy/components/turbomachinery/steam_turbine.py
@@ -117,7 +117,7 @@ class SteamTurbine(Turbine):
     >>> so = Source('source')
     >>> st = SteamTurbine('steam turbine')
     >>> st.component()
-    'turbine'
+    'steam turbine'
     >>> inc = Connection(so, 'out1', st, 'in1')
     >>> outg = Connection(st, 'out1', si, 'in1')
     >>> nw.add_conns(inc, outg)
@@ -135,12 +135,12 @@ class SteamTurbine(Turbine):
     To capture the effect of liquid drop-out on the isentropic
     efficiency, the dry turbine efficiency is specified
     >>> st.set_attr(eta_s=None)
-    >>> st.set_attr(eta_dry_s=0.9, alpha=1.0)
+    >>> st.set_attr(eta_s_dry=0.9, alpha=1.0)
     >>> nw.solve('design')
     >>> round(st.P.val, 0)
     -7009682.0
     >>> round(outg.x.val, 3)
-    0.840
+    0.84
     """
 
     @staticmethod
--- a/src/tespy/components/turbomachinery/steam_turbine.py
+++ b/src/tespy/components/turbomachinery/steam_turbine.py
@@ -16,6 +16,7 @@ from scipy.optimize import brentq
 from tespy.components.component import component_registry
 from tespy.components.turbomachinery.turbine import Turbine
 from tespy.tools.data_containers import ComponentProperties as dc_cp
+from tespy.tools.document_models import generate_latex_eq
 from tespy.tools.fluid_properties import isentropic
 from tespy.tools.fluid_properties import h_mix_pQ
 
@@ -132,7 +133,8 @@ class SteamTurbine(Turbine):
     >>> round(outg.x.val, 3)
     0.821
 
-    To capture the effect of liquid drop-out on the isentropic efficiency, the dry turbine efficiency is specified
+    To capture the effect of liquid drop-out on the isentropic
+    efficiency, the dry turbine efficiency is specified
     >>> st.set_attr(eta_s=None)
     >>> st.set_attr(eta_dry_s=0.9)
     >>> nw.solve('design')
@@ -141,7 +143,6 @@ class SteamTurbine(Turbine):
     >>> round(outg.x.val, 3)
     0.840
     """
-
 
     @staticmethod
     def component():
@@ -156,7 +157,8 @@ class SteamTurbine(Turbine):
         params["eta_dry_s"] = dc_cp(
             min_val=0, max_val=1, num_eq=1,
             func=self.eta_dry_s_func,
-            deriv=self.eta_dry_s_deriv
+            deriv=self.eta_dry_s_deriv,
+            latex=self.eta_dry_s_func_doc
         )
 
         return params
@@ -177,7 +179,8 @@ class SteamTurbine(Turbine):
         else:  # superheated vapour
             dp = inl.p.val_SI - outl.p.val_SI
 
-            # compute the pressure and enthalpy at which the expansion enters the vapour dome
+            # compute the pressure and enthalpy at which the expansion
+            # enters the vapour dome
             def find_sat(frac):
                 psat = inl.p.val_SI - frac * dp
 
@@ -213,7 +216,7 @@ class SteamTurbine(Turbine):
                                 inl.mixing_rule,
                                 T0=inl.T.val_SI
                             )
-            hout = hsat - eta_s * (hsat- hout_isen)
+            hout = hsat - eta_s * (hsat - hout_isen)
 
             # calculate the difference in enthalpy
             dh_isen = hout - inl.h.val_SI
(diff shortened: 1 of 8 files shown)


fwitte:
Good day Tristan, and nice to see your suggestions! I will have a closer look later today and give you feedback on the implementation. Best!

src/tespy/components/turbomachinery/steam_turbine.py
```diff
@@ -0,0 +1,292 @@
...
+    >>> nw.solve('design')
+    >>> round(st.P.val, 0)
+    -7471296.0
+    >>> round(outg.x.val, 3)
+    0.821
+
+    To capture the effect of liquid drop-out on the isentropic
+    efficiency, the dry turbine efficiency is specified
+    >>> st.set_attr(eta_s=None)
+    >>> st.set_attr(eta_dry_s=0.9, alpha=1.0)
```
fwitte, line 140: This should be `eta_s_dry`

src/tespy/components/turbomachinery/steam_turbine.py
```diff
@@ -0,0 +1,292 @@
...
+    >>> from tespy.connections import Connection
+    >>> from tespy.networks import Network
+    >>> from tespy.tools import ComponentCharacteristics as dc_cc
+    >>> import shutil
+    >>> nw = Network(p_unit='bar', T_unit='C', h_unit='kJ / kg', iterinfo=False)
+    >>> si = Sink('sink')
+    >>> so = Source('source')
+    >>> st = SteamTurbine('steam turbine')
+    >>> st.component()
+    'turbine'
```
fwitte, line 122: `'steam turbine'`

src/tespy/components/turbomachinery/steam_turbine.py
```diff
@@ -0,0 +1,292 @@
...
+
+    To capture the effect of liquid drop-out on the isentropic
+    efficiency, the dry turbine efficiency is specified
+    >>> st.set_attr(eta_s=None)
+    >>> st.set_attr(eta_dry_s=0.9, alpha=1.0)
+    >>> nw.solve('design')
+    >>> round(st.P.val, 0)
+    -7009682.0
+    >>> round(outg.x.val, 3)
+    0.840
```
fwitte, line 145: `0.84`, `round()` cuts of trailing zeros

src/tespy/components/turbomachinery/steam_turbine.py
tlmerbecks: Thanks!! I had just found them! sorry very obvious in hindsight

src/tespy/components/turbomachinery/steam_turbine.py
```diff
@@ -0,0 +1,292 @@
...
+        inl = self.inl[0]
+        outl = self.outl[0]
+
+        state = inl.calc_phase()
+        if state == "tp":  # two-phase or saturated vapour
+
+            ym = 1 - (inl.calc_x() + outl.calc_x()) / 2  # average wetness
+            self.eta_s.val = self.eta_dry.val * (1 - self.alpha.val * ym)
+
+            return self.eta_s_func()
```
fwitte, line 191: Question: If we start at saturation or in two phase, we make the assumption, that the expansion path will be wet all the way, correct?

fwitte: P. S.: I kind of dislike the fact, that the function here has a side-effect on a parameter of the component by setting the `eta_s.val` of the component. I would actually suggest, `return self.calc_eta_s() - self.eta_s_dry.val * ...`. Here the reformulation of the isentropic efficiency function does not have any benefits because all residuals will be calculated numerically anyways.

src/tespy/components/turbomachinery/steam_turbine.py
```diff
@@ -0,0 +1,292 @@
...
+                """
+
+        inl = self.inl[0]
+        outl = self.outl[0]
+
+        state = inl.calc_phase()
+        if state == "tp":  # two-phase or saturated vapour
+
+            ym = 1 - (inl.calc_x() + outl.calc_x()) / 2  # average wetness
+            self.eta_s.val = self.eta_dry.val * (1 - self.alpha.val * ym)
```
fwitte, line 189: Should be `eta_s_dry`

src/tespy/components/turbomachinery/steam_turbine.py
```diff
@@ -0,0 +1,292 @@
...
+                                    inl.mixing_rule,
+                                    T0=inl.T.val_SI
+                                )
+                hout = inl.h.val_SI - self.eta_s_dry.val * (inl.h.val_SI - hout_isen)
+
+                # calculate enthalpy of saturated vapour at psat
+                hsat = h_mix_pQ(psat, 1, inl.fluid_data)
+
+                return hout - hsat
+            frac = brentq(find_sat, 1, 0)
```
fwitte, line 216: This makes two expansion sections, right? The first one by finding the pressure, where the expansion enters the dome, then the second one for the rest: The first one applies dry isentropic efficiency, the second one with the correction term?

src/tespy/components/turbomachinery/steam_turbine.py
fwitte: I'll implement this real quick and push it to your branch

src/tespy/components/turbomachinery/steam_turbine.py
```diff
@@ -0,0 +1,289 @@
...
+                hsat,
+                outl.p.val_SI,
+                inl.fluid_data,
+                inl.mixing_rule,
+                T0=inl.T.val_SI
+            )
+            hout = hsat - eta_s * (hsat - hout_isen)
+
+            # calculate the difference in enthalpy
+            dh_bisectioned = hout - inl.h.val_SI
```
fwitte, line 235: I changed this from `dh_isen` to `dh_bisectioned` (we can also make it just sectioned or anything better you may com up with). The reason for this is, that `dh_isen` made me think of the isentropic delta h, but actually it is the real delta h, when factoring in the dry and wet expansion sections, or am I mistaken?

src/tespy/components/turbomachinery/steam_turbine.py
fwitte: I fixed it

src/tespy/components/turbomachinery/steam_turbine.py
fwitte: done

src/tespy/components/turbomachinery/steam_turbine.py
fwitte: I did some benchmark with brentq vs. newton algorithm: `scipy`'s `brentq` is faster by ~factor 3. I will have to look into the fluid property module (and in all other places), where the newton is utilized and see if `brentq` generally performs better :D.

tlmerbecks: Correct, the "wetness" correction is only applied in the two-phase region, for single phase vapour no correction is applied.

So if the vapour is superheated, we first have to find the point where the expansion enters the vapour dome.

tlmerbecks: Regarding Brentq, I performs particularly well if the solution is bracketed, or you know that the function is not necessarily computable for all conditions.

tlmerbecks:
@fwitte I have looked through the changes and agree with the simplification of the residual and the introduction of the `calc_eta_s` function.

I'll have a look at the remaining tests that are still failing later.

Is there anything else you would like to see added/implemented for this?

fwitte:
@tlmerbecks, nothing to implement from your side right now, but I'd have two questions. I have implemented additional tests, and was curious about the following:

- the method does not work, if the outlet is near saturation (see the first test case)
- what happens with isentropic or dry fluids, when the expansion path starts at or above two-phase and the expansion would never cross the two-phase boundary? Or, when the expansion starts in saturated vapor and goes into the gas region?

I guess in both cases we could argue, the method is not applicable here? Or we could make a check, where the inlet and the outlet are (two-phase or gas) and depending on that use one or the other method for expansion? What do you think?

src/tespy/components/turbomachinery/steam_turbine.py
fwitte: Thank you!

fwitte:
And, should we add a reference to the original research, that would be "Baumann, K.: Some Recent Developments In Large Steam Turbine Practice. Engrg. Vol. Ill (1921)" I guess?

tlmerbecks:
Ciao @fwitte ,

I have found a quiet moment :) to answer your questions:

The method is only really applicable if the fluid is water, this is because the correction is entirely based on empirical data on the performance of steam turbines. With this is mind, given the shape of water's vapour dome in the Ts domain it is unnecessary to check for retrograde behaviour.

That being said, perhaps the first check should be whether the phase of the fluid is different between the inlet and outlet. So in the case where the outlet conditions are just above saturation, the fluid enters and exits the turbine as vapour so no correct needs to be applied.

PR #605, opened 2025-02-08
Separate data export from writing to filesystem

fwitte:
This is a breaking change, old style export of tespy networks will not be supported anymore.

- Separate the data export functionality (i.e. writing into a dict of dicts) from writing to the filesystem
- Makes writing to filesystem optional, only when user provides a value for a path
- Returns the dict of dicts
- Implements exerpy export


fwitte:
Two more changes will be implemented before merging this:

- [ ] export dumps to one large json file, network reader will load from that
- [ ] compatibility function to export v0.7.x network to v0.8 style export dump (the large json file)

PR #610, opened 2025-02-12
Implement pareto optimization

fwitte:
This PR builds on the #515 PR because I need both features at once. It will be stale until #515 is merged and published. Only after that, this branch will move forward. It will also break the API, therefore a new major version release will be accompanied with it.


fwitte:
Merged into dev

Issue #600, opened 2025-01-23
Component by-passing/disabling


TristanM-TD:
Ciao Francesco,

Perhaps it would be nice to have the ability to bypass or disable components to be able to switch between different plant configurations more easily.

Bypassing:
If I wanted to investigate the effect of recuperation on the plant performance, I currently have to a) add/remove the recuperator from the network between runs or b) assign a zero duty. With the ability to set a `bypassed` attribute, the component could instead be treated as a `node` with the streams passing through unchanged.

![Image](https://github.com/user-attachments/assets/b9da4721-4e52-43ca-b1a1-599b9c4e45e2)

Disabling:
If I wanted to investigate the effect of only one of my heat exchangers being available as part of an off-design calculation, I would have to a) add/remove the corresponding component from the model or b) artificially assign a very small mass rate (not sure if this would necessarily work). With the ability to set a `disabled` attribute, the branch the component is on could be excluded from the solve.

![Image](https://github.com/user-attachments/assets/99689ea5-60e9-43b3-a34d-63bd22e8d74b)

Not sure if these are the best examples but I hope they convey how this could be useful functiontionality

fwitte:
Ciao Tristan,

that is an excellent idea, I really like it (others already complained about it as well in context of district heating systems), but it is not very easy to generalize. Maybe we can make a start with you bringing up the topic here. A couple of initial thoughts:

- For the bypassing idea, the components could just be replaced by a subsystem interface. One could check in the initialization step of the simulation, whether a component has the `bypass` enabled, and if so, instead of returning its own equations, the ones that would be imposed from a `SubsystemInterface` component. I am not sure, if it makes sense to enable this functionality for any other components than those with n-to-n inlets and outlets (e.g. `HeatExchanger`, `SimpleHeatExchanger`, `Turbine` etc.). This would enforce for all related connections:
  - mass flow and fluid composition equality
  - pressure and enthalpy equality 
- The disabling part might be a little bit more difficult, because setting mass flow to zero is unfortunately numerically challenging (some equations will just not work) and setting it to a very small value might have similar downsides.
[...]

PR #615, opened 2025-02-23
Bypass proof of concept

tlmerbecks:
This PR provides a first proof of concept for the potentially new component bypassing functionality discussed in #600.

### Definition
To summarise, a _bypassed_ `Component` has no affect on the properties of the inlet and outlet streams - effectively the inlet and outlet streams are directly connected to one another.

### Implementation
- adding a new `Component` property called `status` - this would eventually be added to every component. For the time being the default status is "active" and can be set to "bypassed" if a component is to be bypassed.
- adding a new `Component` method called `get_bypass_constraints` - this mirrors `get_mandatory_constraints` method and will eventually have to be implemented specifically for each component
- the `Component` `preprocess` method was adjusted to only add parameter functions if the component is active, otherwise only the bypass constraints are considered.

### Example
In the following example bypassing the `SimpleHeatExchanger ` ignores the parameters set (i.e. defining the duty and pressure ratio)
```python
from tespy.networks import Network
from tespy.components import Sink
from tespy.components import Source
from tespy.components import SimpleHeatExchanger
from tespy.connections import Connection

nw = Network(p_unit='bar', T_unit='C', h_unit='kJ / kg')

source = Source("In")
sink = Sink("Out")
heater = SimpleHeatExchanger("Heater")

c1 = Connection(source, "out1", heater, "in1", "1")
c2 = Connection(heater, "out1", sink, "in1", "2")
nw.add_conns(c1, c2)

c1.set_attr(T=100, p=2, m=1, fluid={"water":1})
heater.set_attr(Q=2e6, pr=1)
heater.set_attr(status="bypass")

nw.solve("design")

nw.print_results()
```

image

### Issues
- It is easy to create conflicts in variable specifications, as often the inlet/outlet conditions of a component are not defined at the `Component` level, but instead for the incomming/outgoing `Connections`...
- Adds unneccessary equations compared to removing the `Component` from the network, although with on-going refactoring in mind this may lose significance.


src/tespy/components/component.py
```diff
@@ -592,14 +601,15 @@ def solve(self, increment_filter):
...
+        if self.get_attr("status") == "active":
+            for data in self.parameters.values():
+                if data.is_set and data.func is not None:
+                    self.residual[sum_eq:sum_eq + data.num_eq] = data.func(
+                        **data.func_params
+                    )
+                    data.deriv(increment_filter, sum_eq, **data.func_params)
 
-                sum_eq += data.num_eq
+                    sum_eq += data.num_eq
```
fwitte, line 612: The dict `self.parameters` should be empty with bypass active anyways, or=

fwitte:
I will do a couple of changes and push them to your branch:

- make it so, that the `solve` call for the components does not need to know, whether or not to bypass the component
- rename to `bypass` (`False` by defalut, but you can just specify `bypass=True`) to a component
- implement another test with a `HeatExchanger` in context of a regenerative preheating

I would like to discuss, to which components it is reasonable to apply this concept. In my opinion, these would be all components with single inlet and outlet or multiple sets of connected inlets and outlets (`HeatExchanger` classes). I would not implement it for combustion or `FuelCell`/`WaterElectrolyzer` components for now, because there it might get more complicated.

What do you think?

src/tespy/components/component.py
fwitte: Ah, it is not. Sorry, I jumped onto this a little bit too fast...

fwitte:
Code for example, here we would have on of these situations, where the user has to undo specifications with bypass activated:

```python
from tespy.networks import Network
from tespy.components import Sink
from tespy.components import Source
from tespy.components import SimpleHeatExchanger, HeatExchanger, Turbine, Pump, Splitter, Merge, Valve
from tespy.connections import Connection

nw = Network(p_unit='bar', T_unit='C', h_unit='kJ / kg')

source = Source("In")
turbine_hp = Turbine("high pressure turbine")
extraction = Splitter("steam extraction")
turbine_lp = Turbine("low pressure turbine")
merge = Merge("condensate merge")
main_condenser = SimpleHeatExchanger("main condenser")
pump = Pump("feed water pump")

preheater = HeatExchanger("preheater")
valve = Valve("condensate throttle")

sink = Sink("Out")


c1 = Connection(source, "out1", turbine_hp, "in1", "1")
c2 = Connection(turbine_hp, "out1", extraction, "in1", "2")
c3 = Connection(extraction, "out1", turbine_lp, "in1", "3")
c4 = Connection(turbine_lp, "out1", merge, "in1", "4")
c5 = Connection(merge, "out1", main_condenser, "in1", "5")
c6 = Connection(main_condenser, "out1", pump, "in1", "6")
c7 = Connection(pump, "out1", preheater, "in2", "7")
c8 = Connection(preheater, "out2", sink, "in1", "8")
nw.add_conns(c1, c2, c3, c4, c5, c6, c7, c8)

c11 = Connection(extraction, "out2", preheater, "in1", "11")
c12 = Connection(preheater, "out1", valve, "in1", "12")
c13 = Connection(valve, "out1", merge, "in2", "13")
nw.add_conns(c11, c12, c13)

c1.set_attr(T=500, p=100, m=10, fluid={"water":1})
c2.set_attr(p=7)

c4.set_attr(p=0.5)
c6.set_attr(p=0.5, x=0)
c7.set_attr(p=100)
c8.set_attr(p=100)

c11.set_attr(m=1)
c12.set_attr(p=7, x=0)

turbine_hp.set_attr(eta_s=0.9)
turbine_lp.set_attr(eta_s=0.88)
pump.set_attr(eta_s=0.75)

nw.solve("design")
nw.print_results()

preheater.set_attr(bypass=True)
c12.set_attr(p=None, x=None)
c8.set_attr(p=None)
nw.solve("design")
nw.print_results()
```

TristanM-TD:
@fwitte thanks for the changes above! 

 > * rename to `bypass` (`False` by defalut, but you can just specify `bypass=True`) to a component

I had previously called it `status` because the grand vision had been to also address the problem of "disabling" components raised in  #600, but perhaps that is a story for another day (i.e once #620 has been completed)


> * I would like to discuss, to which components it is reasonable to apply this concept. 

Regarding roll-out, I agree that it makes most sense for components with single inlet and outlet, or components with multiple inlet/outlet pairs. For all other components, it would be better for the user to define the desired behaviour - perhaps here the `get_bypass_constraints` function should raise a warning (or even an error) to force the user to overwrite the behaviour (e.g. by inheriting from these components).

> * The more I think about this, the more I think, it is a very cool implementation, and the users should be in charge of their problem and understand, what changes in specifications bypass requires for their own problem. What is your opinion?

I think that the bypassing works best when the network performance is driven primarily by the component specifications (i.e. pr, dp, Q, etc.) instead of connection specifications - essentially offdesign calculations. Unfortunately, in my opinion the only way around this would be to either associate connections (or their specifications) with adjacent components, or to specify the inlet/outlet connections directly via the component... The good news is that #620 should make the debugging of such cases more straightforward

fwitte:
Thank you for the feedback!

> I had previously called it `status` because the grand vision had been to also address the problem of "disabling" components raised in #600, but perhaps that is a story for another day (i.e once #620 has been completed)

Okay, then we'll just roll with this implementation for now.

> Regarding roll-out, I agree that it makes most sense for components with single inlet and outlet, or components with multiple inlet/outlet pairs. For all other components, it would be better for the user to define the desired behaviour - perhaps here the `get_bypass_constraints` function should raise a warning (or even an error) to force the user to overwrite the behaviour (e.g. by inheriting from these components).

I will add a warning for the other components then.

> I think that the bypassing works best when the network performance is driven primarily by the component specifications (i.e. pr, dp, Q, etc.) instead of connection specifications - essentially offdesign calculations. Unfortunately, in my opinion the only way around this would be to either associate connections (or their specifications) with adjacent components, or to specify the inlet/outlet connections directly via the component... The good news is that #620 should make the debugging of such cases more straightforward

Or one could create a some kind of mapping, like a per component list of specifications to disable, when then component is flagged bypassed? This would kind of depend also on the style of integration in a use case, so maybe we could discuss a concrete one in a meeting? 

Apart from that, once the warnings are in place I will go ahead and merge. Thank you very much for the support and suggestions.

Issue #612, opened 2025-02-14
Solve of isolated parallel networks


tlmerbecks:
Where two or more hydraulically, thermally and chemically separate networks have been defined under one `Network` - either intentionally or "accidentally" by disabling connecting components (see #600) - the solve should ideally be performed separately. Because,

- if solved as one, the convergence of one depends on the convergence of the other
- the reduction in matrix size speeds up the calculations, although, due to overheads, not all of the potential gains may be realised.

PR #620, opened 2025-03-02
Refactor: presolving based on structural matrix

### LINKED ISSUE #388: FutureWarning: In future versions `DataFrame.to_latex` is expected to utilise the base implementation of `Styler.to_latex`
### Discussed in https://github.com/oemof/tespy/discussions/386

### LINKED ISSUE #465: A method to calculate all partial derivatives of a component's equation numerically might be useful
The designer of a component needs to add equations and the respective derivatives for each desired parameter of that component. There is a possibility to use the `numeric_deriv` function, which calculates a derivative numerically calling the respective equation with a slight change in the specified variable. A list of variables could be added to the parameter information (i.e. the variables relevant for an equation, like inlet enthalpy and pressure, outlet mass flow, ...) and with that list the numeric derivatives would be calculated automatically sparing the designer of the requirement to write the respective `_deriv` method.

### LINKED ISSUE #500: Implementing specified equations into logging
Hi tespy community,

I am currently trying to solve the system of equations for my project and recognized that the specified equations that are listed in the report do not appear in the logging file. To fully understand the calculation I would like to implement that feature into Tespy. 

@fwitte, during the last user meeting, the internet connection was poor, and I couldn't catch everything you mentioned about this issue. I also missed the opportunity to ask about the units in Pygmo, which I'm still curious about. Could we possibly arrange a meeting to discuss this in more detail?

Best regards,
Quentin

fwitte:
Resolve #388
Resolve #500 
Resolve #465 

The changes in this PR will be API breaking, therefore it points to the v0.8 branch.

This is becoming a major refactoring, not all things are finally decided

- All equations can now return a `structure_matrix` entry, which allows to identify linear dependencies in variables in the presolving.
- Linear dependent variables (i.e. a couple of pressure values) will automatically be represented by a singe variable, to which all the linear dependents point.
- The presolver will be able to identify, if users specified
  - more than one value in a set of linear dependent variables
  - have a circular specification in a set of linear dependent variables
  - potentially more things to follow
- The equations collected from connections, components and buses will be indexed. By this the user
  - can investigate, if an equation returns all zero entries for the jacobian
  - can see, which are the equations applied in the model
  - potentially more things to follow
- Placing derivatives in the Jacobian is simplified through wrapper functions

The changes will serve as basis for future changes

- The user can implement custom variables into their problem
- A structural analysis of the problem through a dependency matrix will be possible (are there any separated parts in the matrix, also see #612)
- Power connections (a new connection class, serving for non-material energy flow), with that explicit motor and generator components
- The starting value identification could be refactored to leverage a very similar idea as in the linear dependents finder of the presolving


fwitte:
Update to this:

- There are currently 3 failing tests, which are still expected to fail. The API is changed again by this pr and is incompatible with the API of 0.8/0.7 style exports
- An overall cleanup and aligning of style is required
  - [ ] Define dependents for all equations
  - [ ] Add missing docstrings
  - [ ] Figure out, for which functions using a dedicated `deriv` method is preferred over using the new `dependents` formulation
- Implement easy access to debugging capabilities:
  - Linear dependents finding/presolving
    - [ ] Which equations of which components (and connections) cause circular linear dependencies.
    - [ ] Which specifications of inputs double up on linear dependent variables, e.g. three linear dependent pressure values and two specified (could be direct specification or could be indirect through presolving the state).
- Before start of solving (e.g. in case too many or too few parameters have been provided)
    - [ ] Allow to output all solved (through direct or indirect specification) and unsolved variables and which other variables they represent
    - [ ] Allow to output all presolved equations including the corresponding component/connection and the respective variables
    - [ ] Indicate which equation is dependent on which variable with the dependents mappings
 - During solving
    - In case of a linear dependency, analyze the Jacobian for
      - [ ] rows with all zero and print the respective equations with that issue
      - [ ] combinations of rows that lead to the linear dependency and give the respective set of equations (through Cauchy-Schwarz inequality)
      - [ ] Check in which row there are zero's in for the derivatives, where the dependents martrix would expect a non-zero value
  - [ ] Document all of this in some kind of tutorial

fwitte:
A couple of the above points have already been implemented. There are new points to be added to the list:

- [ ] Improve the data structures/book keeping of which equation goes into which row of the Jacobian. Right now the equation numbers/dependents used for debugging are collected and kept in a dictionary, where the key is the number of row in the overall incidence matrix and the value contains all information. This is not the case for the component/connection/bus residuals and Jacobians. They are still stored with internal counter and not the overall Jacobian. This can potentially lead to the incidence matrix and Jacobian rows not being identically ordered. It should be the overall numbering scheme!
- [ ] It is not entirely clear, how to handle the automatic derivative calculation of single parameters/data containers with multiple equations (num_eq > 1). Ideally, each of the equations would have a separate set of dependents and then each of the equations would be separately derived to all its own partial derivatives. Right now, the derivatives are still put into their places individually for these cases.
- [ ] The `UserDefinedEquation` needs a way to make use of the dependents and users must be forced to specify the dependents by the API. They then can provide the deriv method optionally (otherwise automatic calculation is performed). The dependents also need to be passed to the incidence matrix internally.
- The derivatives for buses are not calculated in the same style as all other derivatives. Since these are subject to API change soon, this will not be changed.

fwitte:
>     * [ ]  Improve the data structures/book keeping of which equation goes into which row of the Jacobian. Right now the equation numbers/dependents used for debugging are collected and kept in a dictionary, where the key is the number of row in the overall incidence matrix and the value contains all information. This is not the case for the component/connection/bus residuals and Jacobians. They are still stored with internal counter and not the overall Jacobian. This can potentially lead to the incidence matrix and Jacobian rows not being identically ordered. It should be the overall numbering scheme!

This is implemented now for components and connections. It is open for the `UserDefinedEquation`. For `Buses` it will continue to work with the old implementation, since there is deprecation planned in this context.

>     * [ ]  It is not entirely clear, how to handle the automatic derivative calculation of single parameters/data containers with multiple equations (num_eq > 1). Ideally, each of the equations would have a separate set of dependents and then each of the equations would be separately derived to all its own partial derivatives. Right now, the derivatives are still put into their places individually for these cases.

This will remain in the current state.

>     * [ ]  A better way has to be found to state variables from a fluid vector as dependents. To simplify a lot of things, the idea is to make all elements of a fluid vector a `ScalarVariable`. Then, this would allow two things:
>       
>       * Identifying dependents just like with all other variables
>       * Calculating and placing partial derivatives of equations towards change in fluid mass fraction into the Jacobian in the same way as is done for ScalarVariables

This will not be changed for now.

>     * [ ]  The `UserDefinedEquation` needs a way to make use of the dependents and users must be forced to specify the dependents by the API. They then can provide the deriv method optionally (otherwise automatic calculation is performed). The dependents also need to be passed to the incidence matrix internally.

Still open

>     * The derivatives for buses are not calculated in the same style as all other derivatives. Since these are subject to API change soon, this will not be changed.

fwitte:
All remaining issues have been solved. Next steps:

- [ ] Update convergence_check methods
- [ ] Fix remaining errors
- [ ] Make compatibility function for v0.8 exports/saves with v0.9 exports/saves (here it may be better to manipulate the serialization/deserialization) so the exports/save files look identical
- [ ] Update component docs on how to create your own components to align with changed (backend) API
- [ ] Update changelog to provide an overview of all (backend) API changes
- [ ] Create a release candidate for 0.9 and publish it to pypi
- [ ] Update the banner on the online documentation to make users test the release candidate and find issues (mostly models, that worked before 0.9 and now stopped working)
- [ ] Write github discussion post announcing same information

fwitte:
In case the number of specifications is incorrect, there should be a best practice workflow on how to still get the equations and the variables to better debug the model in these cases.

PR #661, opened 2025-05-20
Implement calculated dissipative heat losses in pipe

HaSchneider:
**General**

The changes includes a component called "pipeline". It is a pipe component, where dissipative heat losses by forced convection by pipes over the ground and conduction through the soil by burried pipes is calculated. 

**Documentation**

Not documented yet.

**Testing**

Basic test  added to "test_piping.py"

**Questions and Comments**

The component needs several parameters to calculate the heat loss and they are grouped in a 'Q_ohc_group'. But depending on the case (pipes in ground/over ground) slightly different parameters are needed (pipe_depth/wind_velocity). Is there a way to subgroup those parameters?

### DIFF
--- a/src/tespy/components/piping/pipeline.py
+++ b/src/tespy/components/piping/pipeline.py
@@ -12,19 +12,14 @@ SPDX-License-Identifier: MIT
 import math
 #import warnings
 
-#import numpy as np
-
 from tespy.components.heat_exchangers.simple import SimpleHeatExchanger
 from tespy.components.component import component_registry
-#from tespy.tools import logger
-#from tespy.tools.data_containers import ComponentCharacteristics as dc_cc
 from tespy.tools.data_containers import ComponentProperties as dc_cp
 from tespy.tools.data_containers import GroupedComponentProperties as dc_gcp
 from tespy.tools.data_containers import SimpleDataContainer as dc_simple
 from tespy.tools.document_models import generate_latex_eq
-#from tespy.tools.fluid_properties import s_mix_ph
-#from tespy.tools.fluid_properties.helpers import darcy_friction_factor as dff
-#from tespy.tools.helpers import convert_to_SI
+
+from tespy.tools.helpers import convert_to_SI
 from tespy.tools.fluid_properties.wrappers import CoolPropWrapper
 
 @component_registry
@@ -180,8 +175,7 @@ class Pipeline(SimpleHeatExchanger):
     >>> inc.set_attr(fluid={'water': 1}, m=100, T=300, p=40)
     >>> pi.set_attr(pr=0.95, Tamb = 20, L=1000, D='var',  ks=4.57e-5,
     >>>                     insulation_m=0.1 ,insulation_tc= 0.035, pipe_thickness=0.003,material='Steel', 
-    >>>                     environment_media= 'air', wind_velocity = 0.1,
-    >>>                     #environment_media= 'dry soil', pipe_depth = 5,
+    >>>                     environment_media= 'air', wind_velocity = 0.1,        
     >>>             ) 
     >>> nw.solve('design')
     >>> nw.save('tmp')
@@ -190,6 +184,13 @@ class Pipeline(SimpleHeatExchanger):
     @staticmethod
     def component():
         return 'Pipeline'
+    
+    def preprocess(self, num_nw_vars):
+        self.air= CoolPropWrapper('air')
+
+        super().preprocess(num_nw_vars)
+
+        self.Tamb.val_SI = convert_to_SI('T', self.Tamb.val, self.inl[0].T.unit)
 
     def get_parameters(self):
         parameters=super().get_parameters()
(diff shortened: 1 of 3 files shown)


tests/test_components/test_piping.py
```diff
@@ -83,3 +84,26 @@ def test_Pipe(self):
...
+        # parameter specification
+        self.c1.set_attr(fluid={'H2O': 1}, m=10, p=10, T=220)
+        
+        instance.set_attr(pr=0.99, Tamb = 20, L=1000, D=0.2, 
+                       insulation_m=0.1 ,insulation_tc= 0.035, pipe_thickness=0.002,
+                       material='Steel', 
+                       environment_media= 'air', wind_velocity = 2,
+            ) 
+        self.nw.solve('design')
+        self.nw._convergence_check()
```
fwitte, line 102: After this we should make a check for the correctness of the test case, do you have any reference values?

tests/test_components/test_piping.py
```diff
@@ -83,3 +84,26 @@ def test_Pipe(self):
...
+        instance.set_attr(pr=0.99, Tamb = 20, L=1000, D=0.2, 
+                       insulation_m=0.1 ,insulation_tc= 0.035, pipe_thickness=0.002,
+                       material='Steel', 
+                       environment_media= 'moist soil', pipe_depth = 5,
+            ) 
+        self.nw.solve('design')
+        self.nw._convergence_check()
```
fwitte, line 109: Same here

src/tespy/components/piping/pipeline.py
```diff
@@ -0,0 +1,378 @@
...
+            '''
+            Gnielinski, V.: Berechnung mittlerer Wärme- und Stoffübergangskoeffizienten an laminar
+            und turbulent überströmten Einzelkörpern mithilfe einer einheitlichen Gleichung.
+            Forsch. Ing.-Wes. 41(5), 145-153 (1975)
+            '''
+            Re = (
+                self.wind_velocity.val * math.pi / 2
+                * (diameters[1] + self.insulation_m.val * 2)
+                / self.air.viscosity_pT(101300, self.Tamb.val_SI + 273)
+                * self.air.d_pT(101300, self.Tamb.val_SI + 273))
```
fwitte: Why are these adding 273? The values should already be SI.

src/tespy/components/piping/pipeline.py
```diff
@@ -0,0 +1,378 @@
...
+        if i.h.is_var:
+            self.jacobian[k, i.h.J_col] = self.numeric_deriv(func, 'h', i)
+        if o.h.is_var:
+            self.jacobian[k, o.h.J_col] = self.numeric_deriv(func, 'h', o)
+        if i.p.is_var:
+            self.jacobian[k, i.p.J_col] = self.numeric_deriv(func, 'p', i)
+        if o.p.is_var:
+            self.jacobian[k, o.p.J_col] = self.numeric_deriv(func, 'p', o)
+        if self.D.is_var:
+            self.jacobian[k, self.D.J_col] = self.numeric_deriv(func, 'D', None)
```
fwitte, line 391: I added the diameter as potential variable. I can see, that there is a case for identifying the pipe diameter while simultaneously imposing a pressure drop and a heat loss to the ambient.

src/tespy/components/piping/pipeline.py
```diff
@@ -0,0 +1,378 @@
...
+from tespy.tools.data_containers import SimpleDataContainer as dc_simple
+
+from tespy.tools.helpers import convert_to_SI
+from tespy.tools.fluid_properties.wrappers import CoolPropWrapper
+
+
+@component_registry
+class Pipeline(SimpleHeatExchanger):
+    r"""
+    Pipeline for modeling steam networks.
```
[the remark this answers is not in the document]
HaSchneider: No, it should work for all types of flows, regardless of the components and aggregate state. Compared to the pipe/SimpleHeatExchanger it calculates the heat transfer by a heat transfer coeficient depending on equipment and external parameters instead of a fixed definition like in the kA group. Its actually more like an extension of the pipe component. For now I created a separat component for it, because I was not sure if it will support all parameter groups of the pipe component. But probably it would be better to directly merge it in the pipe component?

src/tespy/components/piping/pipeline.py
HaSchneider: Yes, you are right! Its removed now.

fwitte:
> The commit [a7e3b41](https://github.com/oemof/tespy/commit/a7e3b41d6187f598b2475d9143a8bfa78da45c58) raises an error for me:
> 
> ```
> AttributeError: 'Pipeline' object has no attribute 'dp'
> ```
> 
> I uncommented it in my last commit, for what is it needed?

That is for postprocessing. The `SimpleHeatExchanger` postprocessing implements `kA`, which I deemed not relevant for this component. My thought was: Since one would specify so many details on the component here, you would not use the `kA` equation. For that reason, I removed the `kA` parameter also from the parameter list and then overwrote the postprocessing method.

`dp` is the pressure drop on the component. It should be available through the component's parameters as they are inherited from the `SimpleHeatExchanger` (or `Pipe`).

src/tespy/components/piping/pipe.py
```diff
@@ -448,7 +443,7 @@ def ohc_subsurface_group_func(self):
 
         return (
             i.m.val_SI * (o.h.val_SI - i.h.val_SI)
-            + 1 / R_soil * self._calculate_td_log()
+            + 1 / R_soil * self._calculate_td_log() * self.L.val
```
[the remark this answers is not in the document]
fwitte: Got it, thank you! I'll release tomorrow:)

fwitte:
Completed! Thank you for your contribution @HaSchneider

PR #713, opened 2025-07-24
Connectors' number of PowerConnection upadted in _to_exerpy

sertomas:
The _to_exerpy function in the PowerConnection class now assigns the correct number to each connector, considering the other connections of PowerBus, Compressor, Pump or Turbine classes.


fwitte:
Hi @sertomas 

thank you for the PR, I merged it and published a release.

Best

PR #774, opened 2025-09-18
Fix scaling issues for humid air

fwitte:
Currently there is a persisting issue with calculations at saturation, because it is not possible to calculate with partial pressure of water at temperature, when it corresponds to the saturation pressure of that temperature. Checking for saturation before with sufficient precision is now done based on relative deviation.

When taking bigger steps in change of liquid phase the temperature tends to either go towards 0 °C or critical point temperature. Here the derivatives will become zero because the change of enthalpy then does not affect calculated temperature and/or liquid mass fraction anymore. Ideally, a way to relax the equation can be found.


fwitte:
The tests here are only failing due to a broken link in the docs (which might go online soon again). Therefore I will merge.

Issue #808, opened 2025-10-17
Restructure the documentation to make orientation better


fwitte:
The documentation should be restructured a bit to make the orientation inside the docs easier, draft:

- Landing page with an overview similar to the landing page on the tutorial or module overview and there link to
  - Installation instructions
  - Introductory tutorials and educational (external) resources 
  - User manual with basic information (model structures, specification of parameters, selection of units, accessing results, postprocessing, debugging)
  - Advanced tutorials
  - User manual with advanced information (subsystems, user defined equations, adjusting existing components or developing custom ones, contribute to the development)
  - API documentation and other topics

For the API documentation, a better entry point could be chosen as well, the idea is:

- Create a tabular overview on Connection, PowerConnection and all Component type classes, the table will
  - list specification options per class
  - link to the respective API docs of the component class and of the methods associated with the parameters

maltefritz:
Hey everyone,

I think the idea of making the documentation even clearer is great. I also really like the suggestion for structuring the API.

In my opinion, it would also be helpful to place the simplified architecture from the TESPy poster in a suitable location. In addition, explanatory illustrations throughout would also help beginners get started with TESPy.

One more thing: the announcement for the next online TESPy meeting is incorrect.

PR #817, opened 2025-10-24
New tutorial for how to contribute a component.

maltefritz:
After the hackathon at the TESPy community meeting, I created a tutorial on 'how to integrate a new component'. It's about the polynomial compressor with cooling. I tried to write down the workflow from the meeting. However, since this was an ongoing process, it was more difficult than I thought. That's why I would really appreciate a review.

I placed the tutorial as a dropdown under “how can I contribute” in TESPy. To do this, I had to change the location of the _how_-file. If this is not necessary, it could be changed.

At the end of the documentation, I added a note on “how to write a docstring in TESPy” and “how to write tests.” I deliberately did not edit this so as not to make the tutorial too long. Furthermore, I think it would be useful to provide brief instructions on these two topics as well, in order to make the transition from user to developer as easy as possible for many people.


fwitte:
Thank you very much for the contribution @maltefritz! I will have a look at this in the upcoming weeks. At first glance, I'd like to have this moved to somewhere else in the table of contents, because it seems rather difficult to find. Since a general restructuring was suggested (#808) I'd think we can make the restructuring and then merge this contribution into that :).

PR #820, opened 2025-10-25
Restructure the documentation to make orientation easier

fwitte:
Resolve #808 
Resolve #853 

Here we can develop a new version of the landing page. Right now, the idea is to have a two-layer approach:

- First layer: Landing page linking to the second layer
- Second layer: A sublanding page linking to the respective subsections

Right now there are still template images in the grid, which will be replace by better ones. The first idea would include 9 cards to start with:

- Introduction, Installation, Tutorials
- High-level overview on Network, Components, Connections
  - Specification of parameters
  - Retrieving results/getting parameters  
  - Tabular overview of inputs, outputs, used equations
  - ..?
- Advanced Tutorials (links to current overview of advanced tutorials section)
- Community and Support (direct link to current community and support page)
- Advanced Features (links to overview of UserDefinedEquations, fluprodia integration, customization of components, adding new components)
- API and changelog
- Example projects, Benchmarks, Literature (links to a subpage redirecting to the respective information)
- GitHub Repository
- Contribute to development

Any thoughts on this @25Nik, @maltefritz, @jfreissmann, @TristanM-TD?

You can always view current state of draft here: https://tespy--820.org.readthedocs.build/en/820/

PR #838, opened 2025-11-13
Add FAQ entries and suggestions for answers.

jfreissmann:
Help fix Issue #808 by adding some entries to the new FAQ section in alignment with the to-do-list in PR #820.

I am very much open to more FAQ entries and happy get suggestions concerning any changes to the entries or their order within the FAQ section.

### DIFF
--- a/docs/knowledge_center/faq.rst
+++ b/docs/knowledge_center/faq.rst
@@ -215,9 +215,16 @@ Frequently Asked Questions
 
     .. dropdown:: How can I apply Second Law of thermodynamics analysis methods?
 
-        - Introduce exergy analysis and how it can help
-        - `exerpy <https://exerpy.readthedocs.io/en/latest/>`__
-        - Reference other methods like entropy analysis etc.
+        TESPy has an integrated exergy analysis feature. This was developed
+        originally in two publications: :cite:`Witte2022,Hofmann2022`. The
+        feature automatically calculates the exergy of all streams in your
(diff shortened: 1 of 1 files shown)


docs/knowledge_center/faq.rst
```diff
@@ -3,43 +3,131 @@
...
+
+            fig, ax = plt.subplots(figsize=(10, 6))
+
+            ax.plot(heat, T_hot, "o-", color="red")
+            ax.plot(heat, T_cold, "o-", color="blue")
+
+            ax.set_ylabel("temperature in K")
+            ax.set_xlabel("heat transferred in MW")
+
+            plt.show()
```
fwitte: It would be nice if theres were testable. Maybe we can do the following:

```python
>>> from tespy.tools import HeatExchangerTestNetwork (or from where-ever)
>>> nw = HeatExchangerTestNetwork()
>>> heatex = nw.get_comp("heatex")
>>> nw.solve("design")
>>> ...
>>> [the plotting code]
```
Lines with >>> in front will automatically be tested by the testing pipeline, so we would be notified if anything breaks with this example.

fwitte: True!

But: I kind of like the idea to just grab a TestNetwork from the testnetworks module whenever I need one for fiddling around... Once you get used to them, these could be very useful for quick testing or showing things. You do not need to do all the imports etc..

I'd say: You decide what you prefer :)

PR #855, opened 2025-11-30
Implement an example for pinch analysis through coupling `tespy` with `pina`

n-arth:
**Pull request workflow**

Resolves #ISSUE_NUMBER

I added the aspects from pina for pinch analysis and additional plotting functions. I also added comments for still needed functions as well as possibilities to connect to tespy. By now it can be seen as a seperated set of functions.

- [ x] Implement unit tests for the changes --> done as an example, as the implementation is currently by plotting
- [ x] For a new feature: Implement an example and a system test --> example as said above

I need help for those two aspects
Additionally pina 0.1.1 is required. Add this the requirements


fwitte:
@n-arth, thank you for progressing this. I think we could make it some kind of tutorial soon. Couple of thoughts:

- I would like to move the pinch class out of the package for the moment, to not have it as hard (to be maintained) dependency
- Instead I suggest we transpose the files into some kind of notebook and move it to the advanced tutorials section. There with a bit of text, some flowsheets and diagrams one can go step by step through the modeling. Essentially similar to this https://tespy.readthedocs.io/en/dev/advanced_tutorials/co2_cycle.html#tutorial-co2-refrigeration-label or this https://tespy.readthedocs.io/en/dev/advanced_tutorials/heat_pump_exergy.html#tutorial-heat-pump-exergy-label.

I can make a draft for the notebook and the flowsheets and will push it as soon as I find the time for it

n-arth:
Hi @fwitte, thank you for your feedback.
I guess moving the whole class into an example would be good. That way, multiple analysis can be set up in parallel to e.g. compare integrations and also to further compare the different heat exchangers and limitatios within pinch analysis. 

On the long run I would suggest looking for a way to relace pina with OpenPinch. For the meantime, copying the class from the example should work.

What do you think?

PR #860, opened 2025-12-12
Fix font used in 'About TESPy' thumbnail.

jfreissmann:
The current thumbnail used for the "About TESPy" section in the restructured and improved documentation uses the wrong font. This PR fixes this.


fwitte:
Perfect, thank you!

PR #941, opened 2026-04-12
Export 'vol' as keys for specific volume for fluprodia plotting

### LINKED ISSUE #432: [BUG?] v prop in HEX cold side
Hi,
first of all, I thank you for the great work you are doing. 
I just started with the library and it seems very well thought out and structured. 
Could you please  check the properties present in the connections of a Heatexchanger in the cold branch? In my opinion p, T, h are correct, v is wrong (other properties I have not checked). Thank you

fwitte:
Resolves #432

- [x] Update the documentation
- [x] Update the changelog at docs/whats_new/*

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ["3.12"]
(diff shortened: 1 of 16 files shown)


fwitte:
The tests are failing because they do not yet know about latest fluprodia

PR #982, opened 2026-05-12
Add modeltemplate support for parameters referencing multiple components

HaSchneider:
There are cases when a model template parameter set multiple component parameters. For instance: efficiencies of all pumps in a network. 
As far as I know, right now a parameter for each component need to be defined for that. 

My suggestion is that you can define a list in the lookup method of the model template, containing all parameter references that pertain to the parameter. 

Todos:
- [ ] Implement unit tests for the changes
- [ ] For a new feature: Implement an example and a system test

### DIFF
--- a/src/tespy/models/template.py
+++ b/src/tespy/models/template.py
@@ -71,14 +71,24 @@ class ModelTemplate():
         if parameter not in self.parameter_lookup:
             raise KeyError(f"'{parameter}' is not in parameter_lookup.")
         mapped = self.parameter_lookup[parameter]
-        if isinstance(mapped, dict):
-            if "get" not in mapped:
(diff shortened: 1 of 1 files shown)


HaSchneider:
Adding support for parameters referencing factors of Ref classes would be also nice. But I think this would need some kind of Ref registry with unambiguous naming of all Ref classes.

HaSchneider:
Ah, I just recognized, that you can also define those dictionaries in the parameter_lookup function, which set/get values can be any callable and thereby offers a flexible option to set multiple parameters with the same value or set reference flows here. Then I think this pull request is not required. But probably it can be explained more briefly in the documentation ;-)

Issue #976, opened 2026-05-08
`zeta` value description is misleading and a real `zeta` value in combination with `D` might be useful


fwitte:

fwitte:
Change has been implemented in version 0.10

PR #989, opened 2026-05-17
Rename kA to UA and zeta to zeta_d4

### LINKED ISSUE #979: Transition from `kA` to `UA` for all components of type `HeatExchanger`
All `kA` specifications will be renamed to `UA` or replaced:

- `SimpleHeatExchanger`, `HeatExchanger`, `Condenser`, `Desuperheater`, `ParallelFlowHeatExchanger`: essentially just a rename to `UA`
- `SectionedHeatExchanger`, `MovingBoundaryHeatExchanger`: remove `kA` entirely

fwitte:
Resolves #979 
Prepares #976 change for version 0.11

This PR will also add a new consistent parameter `lmtd` for all heat exchanger types which is equal to `Q/UA`

- [x] Implement unit tests for the changes
- [x] Update the documentation
- [x] For a new feature: Implement an example and a system test
- [ ] (Optional) update the CITATION.cff file, if you want to be included in
      the citable contributor list


fwitte:
Everything changed in src tests still all pass (except exerpy related ones), so green light here.

fwitte:
Last missing piece is a unit test that verifies the conversion of the parameters. Right now the tests implicitly do because all the test still use the old zeta/kA specifications. But this will also be changed, so there must be a test that checks if specification of a deprecated parameter which will be renamed will correctly specify the renamed parameter and its result will correctly be passed back to the deprecated parameter. This is to be checked for

- the value itself
- is_set
- is_var

PR #990, opened 2026-05-17
Implement example of bulk setting pressure drops

### LINKED ISSUE #983: Give examples how the callables can be used in set or get parameters of ModelTemplate
> Ah, I just recognized, that you can also define those dictionaries in the parameter_lookup function, which set/get values can be any callable and thereby offers a flexible option to set multiple parameters with the same value or set reference flows here. Then I think this pull request is not required. But probably it can be explained more briefly in the documentation ;-)

_Originally posted by @HaSchneider in https://github.com/oemof/tespy/issues/982#issuecomment-4430270429_

fwitte:
Resolves #983 

- [X] Update the documentation

### DIFF
--- a/docs/integration/workflows.ipynb
+++ b/docs/integration/workflows.ipynb
@@ -126,7 +126,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "4f9138b2",
+   "id": "409f7de1",
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


fwitte:
What do you think about this @HaSchneider?

fwitte:
https://tespy--990.org.readthedocs.build/en/990/integration/workflows.html#heat-pump-with-two-refrigerant-cycles

HaSchneider:
Thanks, I think the usage is clearer now!

PR #1001, opened 2026-06-03
Calc_sections() modified in HeatExchanger base.py (Units updated based on user-specified units)

Flefays:
I was using the model template, which is allowing to add labels externally but, still plot the default units.

I import Units in heatexchanger base.py and convert it to the user specification.

I hardwritten the default units regarding units.py maybe this point could be automatised as well.

For example I wrote "W" for the heat, so if you want to modify the default unit of tespy ( Of course not, but in the idea), this has to be updated along. Or maybe a default_unit command exist to replace this part.

Now my code using the model template Q-T function plot in my specified unit perfectly the Q-T (verified).

- [ x] (Optional) update the CITATION.cff file, if you want to be included in
      the citable contributor list


fwitte:
Hi @Flefays, thank you for the suggestion. Let me understand: you want the QT diagram data and axes lables to automatically show in the user specifications for the network units?

Flefays:
> Hi @Flefays, thank you for the suggestion. Let me understand: you want the QT diagram data and axes lables to automatically show in the user specifications for the network units?

Exactly. The Q-T diagrams using Calc_sections() are displaying the default units. So locally I imported the tespy tool unit in HeatExchanger base.py and  I converted  the units to the user specifications for the network units. Allowing the Model template Q-T function to plot °C if I'm working in °C. 

However I hardcoded the default units copying the ones in units.py, because, I didn't found a command.

It works like this, but if you change in the source code the heat default units from "W" to "apples", then my modification is caduc, because I hard coded the heat default units was "W". For example the temperature difference is by default in delta_degC, then I wrote:

   td_log_per_section = _UNITS.ureg.Quantity(td_log_per_section, "delta_degC").m_as(self.td_log.unit)

fwitte:
The problem with this change is, that the method is also internally used and for that it will need to return SI units. I think I will implement a `calc_sections_SI` method which can be used internally, and this method can return results in user units. However, since it always returned SI units in the past we cannot make that behavior change without prior warning. So the procedure is:

- introduce `calc_sections_SI` method
- introduce temporary `calc_sections_user_units` method
- issue a FutureWarning in `calc_sections`, that the behavior will change and it will no longer return SI unit values in the future (version 0.11 or 1.0) and users should revert to `calc_sections_SI` for SI results and `calc_sections_user_units` temporarily until the behavior changed for user units.

fwitte:
I think this is going to be a bit of a bigger refactor, but one that also has some other nice additions:

- I will add array properties for components to the results of the sections can sit in component attributes both in SI and in user units
- calc_sections will warn the user how to properly use these parameters in the future and entirely removed in next major release

fwitte:
@Flefays, thank you for bringing up the topic. It now went a very different direction from the first idea, but I think this is much cleaner. You can check in the dev branch of tespy (or your dev branch since I pushed to there) in the workflows.ipynb in the docs folder how the new method is utilized internally. You can after solving explicitly access the section states, which will be available in SI (.val_SI) and in network untis (.val)

Flefays:
@fwitte, thank you for your consideration.

PR #1021, opened 2026-07-02
Feature method to convert network to subsystem

HaSchneider:
As described in #1020 this implements a method for the Network class which returns a subsystem class implementation of the corresponding network.

A new subsystem is created and all sinks and sources are transformed in subsystem interfaces. All connections and the corresponding components are converted, only those to/from sink/sources are reconnected to subsystem interfaces.

This doesnt have to be a method of Network class, it could also be implemented as normal function in subsystem.py for example. 

To do:

- [x] Add PowerSources and PowerSinks
- [x] Refactor code
- [x] Implement unit tests for the changes
- [ ] Implement an example and a system test
- [x] Update the changelog at docs/whats_new/*


fwitte:
Thank you for the contribution @HaSchneider, will look into this and give you feedback soon!

fwitte:
As it could be possible to have subsystem-internal sources/sinks, I would like to suggest the following change:

- expose all sources and sinks by default
- allow user to specify a list of sources and sinks to be exposed

I would also add/transfer the properties of the original connections to the interface connections so starting values can be preserved, if the original Network was solved already.

Other than that, I am thinking if the Subsystem/Network architecture needs to be implemented in a different way, there are a couple of thoughts that come to my mind:

- What about a Subsystem in a Subsystem in a Subsystem ...
- Serialization/Deserialization of Subsystem instances (right now the components and connections are flattened by prefixing their labels with the subsystem label) and then serialized/deserialized through Network
- Allow to solve Subsystem independently of Network or other way round by fixing what ever the specifications/boundary conditions are at the interfacing connections.

Any thoughts about it, since you are working with Subsystems frequently? This does not need to be included in this PR but it let to me thinking about it :D

HaSchneider:
That are good suggestions. I adjusted the method, so that all connections are copied instead of creating new conenctions. All connection properties should be kept now. Only the source/targets are now adjusted accordingly to the subsysteminterface. 
I also added an argument to exclude sinks/sources from conversing them to subsystem interfaces. 

Regarding your other suggestions:
- "nested subsystems" are already possible, arent they? 
- right now subsystems are "only" grouping network parts, but without much functionalitites, right? So nice features I could imagine are (just a brainstorming):
    - Solving subsystem independly as you already mentioned. I think that would make large networks easier to maintain and debug. And would probably allow an implementation of an hybrid of sequential modular and simultaneous solving. 
    - import/export of subsystems: for that a dedicated serialization/deserialisation of subsystems might be nice. So that only the subsystem can be exported and imported without the need of import/export the whole network.  
    - combine modeltamplate and subsystems: 
        - allow subsystem specification in modeltemplate classes with corresponding method to converse/export the modeltemplate to a subsystem with defined interface exceptions and starting values etc. 
        - plotting subsystems in a model template seperatly or in seperat colors

fwitte:
> That are good suggestions. I adjusted the method, so that all connections are copied instead of creating new conenctions. All connection properties should be kept now. Only the source/targets are now adjusted accordingly to the subsysteminterface. I also added an argument to exclude sinks/sources from conversing them to subsystem interfaces.

Nice, thank you! I will have a look at the changes. All the other topics can go to a different issue/pr.

>     * "nested subsystems" are already possible, arent they?

It is, but the inherit logic of which subsystem belongs to where goes missing due to the flattening of the labels. So a structured registry in the Network could be helpful.

>       * Solving subsystem independly as you already mentioned. I think that would make large networks easier to maintain and debug. And would probably allow an implementation of an hybrid of sequential modular and simultaneous solving.

Yes, that is exactly what I am thinking about as well. Also being able to put parts of Network in "shutdown" mode.

>       * import/export of subsystems: for that a dedicated serialization/deserialisation of subsystems might be nice. So that only the subsystem can be exported and imported without the need of import/export the whole network.

also agree

>         * allow subsystem specification in modeltemplate classes with corresponding method to converse/export the modeltemplate to a subsystem with defined interface exceptions and starting values etc.
>         * plotting subsystems in a model template seperatly or in seperat colors

also agree

src/tespy/networks/network.py
```diff
@@ -4043,7 +4047,97 @@ def _export_components(self):
...
+                for inner_key in inner_dict.keys()}
+                out_i=1
+                in_i=1
+                heat_out_i=1
+                heat_in_i = 1
+                power_out_i=1
+                power_in_i = 1
+                
+                for conn in args:
+                    self.conns[conn.label] = copy.copy(conn)
```
fwitte: I think the copying/plain reusing of connections could break things if a model is exported more than once. There will be still some objects in the copied Connection instances which point to the original objects, e.g. fluidwrappers etc. So maybe it is actually better to serialize and de-serialize instead.

src/tespy/networks/network.py
```diff
@@ -4043,7 +4047,97 @@ def _export_components(self):
                 components[c].update(cp._serialize())
 
         return components
-
+    
+    def convert_to_subsystem(self, subname, interface_exceptions=[]):
```
fwitte: Maybe this could live in the `Subsystem` class instead. So we could have something like this:

- `Subsystem.from_network()`
  - calls `Network.export()`, your logic to transform the ports and then `Subsystem._deserialize()` to deserialize the `Network.export()`
- `Subsystem.to_network()`
  - calls `Subsystem._serialize()`, logic to transform `SubsystemInterface` to `Source`/`Sink` etc., `Network.from_dict()`
- `Subsystem._deserialize()`
- `Subsystem._serialize()`

fwitte:
Let's maybe have a call to discuss this topic in detail @HaSchneider, do you have time Monday next week?

src/tespy/networks/network.py
HaSchneider: Yes, that is correct. I tried also deepcopy but that results in the follwoing error:
```python
TESPyComponentError: Component name in subsystem is not unique
```

src/tespy/networks/network.py
HaSchneider: Yes, this might be a better location. Currently the type of the returned Subsystem is:
```python
tespy.networks.network.Network.convert_to_subsystem.<locals>.sub_network
```
:laughing:

HaSchneider:
The subsystem has now a clasmethod `from_network(label, nw)` which returns a subsystem created from the label.

There is one issue that might come up here: When a proper configured network is converted to a subsystem, using that subsystem in another network will most likely lead to an overdetermined system.

fwitte:
Thank you for the update, we could change it so the subsystem can optionally be created "clean" = without any specifications.

fwitte:
Okay, I think this is mostly ready now:

- there is `interface_map` giving you the information, which port at the `SubsystemInterface` now represents which component (`Source`/`Sink`) from the original network
- all boundary conditions are retained as they were specified before exporting the `Network`

HaSchneider:
Nice, my examples works with that.

> Thank you for the update, we could change it so the subsystem can optionally be created "clean" = without any specifications.

Is this also already possible somehow? Would expect that the method 'Subsystem.from_network()' has an argument for that.

fwitte:
> Is this also already possible somehow? Would expect that the method 'Subsystem.from_network()' has an argument for that.

I wanted to make a separate method for this, that could be done for `Network` instances in general (see #1038):

```python
nw.clear_specifications()
specs = nw.save_specifications()
nw.restore_specifications(specs)
```
That could integrated with the `from_network` method then

HaSchneider:
The following error occurs when an inititated subsystem from an network is used in multiple networks. Probably this error can be handled with a more specific error message:
`ValueError: Cannot operate with Unit and Unit of different registries.`

````python
def _compressor_network():
    nw = Network(iterinfo=False)
    nw.units.set_defaults(pressure_difference="bar", temperature="degC")
    so = Source("air inlet")
    cp = Compressor("compressor")
    si = Sink("air outlet")
    ps = PowerSource("grid")

    c1 = Connection(so, "out1", cp, "in1", label="1")
    c2 = Connection(cp, "out1", si, "in1", label="2")
    p1 = PowerConnection(ps, "power", cp, "power", label="p1")
    nw.add_conns(c1, c2, p1)

    c1.set_attr(fluid={"air": 1}, T=20, p=1, m=1)
    c2.set_attr(p=5)
    cp.set_attr(eta_s=0.8)
    return nw
def test_subsystem_from_network():
    nw = _compressor_network()
    nw.solve("design")
    nw.assert_convergence()

    sub = Subsystem.from_network("compressor block", nw)


    nw2 = Network(iterinfo=False)
    nw2.units.set_defaults(pressure_difference="bar", temperature="degC")
    so = Source("source")
    si = Sink("sink")
    ps = PowerSource("power source")

    c_in = Connection(so, "out1", sub.inlet, "in1", label="c_in")
    c_out = Connection(sub.outlet, "out1", si, "in1", label="c_out")
    p_in = PowerConnection(ps, "power", sub.inlet, "power_in1", label="p_in")
    nw2.add_conns(c_in, c_out, p_in)
    nw2.add_subsystems(sub)
    nw2.solve("design")

    nw3 = Network(iterinfo=False)
    nw3.units.set_defaults(pressure_difference="bar", temperature="degC")
    so = Source("source")
    si = Sink("sink")
    ps = PowerSource("power source")

    c_in = Connection(so, "out1", sub.inlet, "in1", label="c_in")
    c_out = Connection(sub.outlet, "out1", si, "in1", label="c_out")
    p_in = PowerConnection(ps, "power", sub.inlet, "power_in1", label="p_in")
    nw3.add_conns(c_in, c_out, p_in)
    nw3.add_subsystems(sub)
    nw3.solve("design")

test_subsystem_from_network()
````

fwitte:
The error actually also occurs when adding the same component instance to different networks. I think making it fail with a clear error message is the best way to go, because otherwise we need to deal with the fact, that connection and component properties can be modified from two different network instances.

fwitte:
Okay, missing docs update only. I will merge this, and #1047 afterwards

PR #1025, opened 2026-07-12
minor changes in documentation of advanced HX tutorial

n-arth:
- changed label to match used HX model as discussed
- changed one word in the description, as it makes clarification easier while scanning through the page


fwitte:
Thank you for the fix @n-arth

